PR #9, opened 2020-06-09
Add more tests

AKS1996:


AKS1996:
@matbesancon am i checking test exception right for non-convex QP?

test/diff_model.jl
```diff
@@ -93,3 +93,44 @@ end
...
+                quad_terms, 
+                MOI.ScalarQuadraticTerm(Q[i,j], x[i], x[j])
+            )
+        end
+    end
+
+    objective_function = MOI.ScalarQuadraticFunction(
+                            MOI.ScalarAffineTerm.(q, x),
+                            quad_terms,
+                            0.0
```
matbesancon, line 121:
```suggestion
                            0.0,
```

test/diff_model.jl
```diff
@@ -93,3 +93,44 @@ end
...
+    model = MOI.instantiate(OSQP.Optimizer, with_bridge_type=Float64)
+    x = MOI.add_variables(model, 2)
+
+    # define objective
+    quad_terms = MOI.ScalarQuadraticTerm{Float64}[]
+    for i in 1:2
+        for j in i:2 # indexes (i,j), (j,i) will be mirrored. specify only one kind
+            push!(
+                quad_terms, 
+                MOI.ScalarQuadraticTerm(Q[i,j], x[i], x[j])
```
matbesancon, line 113:
```suggestion
                MOI.ScalarQuadraticTerm(Q[i,j], x[i], x[j]),
```

test/diff_model.jl
```diff
@@ -93,3 +93,44 @@ end
...
+                            0.0
+                        )
+    MOI.set(model, MOI.ObjectiveFunction{MOI.ScalarQuadraticFunction{Float64}}(), objective_function)
+    MOI.set(model, MOI.ObjectiveSense(), MOI.MIN_SENSE)
+
+    # add constraint
+    MOI.add_constraint(
+        model,
+        MOI.ScalarAffineFunction(MOI.ScalarAffineTerm.(G[1, :], x), 0.0),
+        MOI.LessThan(h[1])
```
matbesancon, line 130:
```suggestion
        MOI.LessThan(h[1]),
```

test/diff_model.jl
```diff
@@ -93,3 +93,44 @@ end
...
+    )
+
+    diff = diff_model(model)
+
+    @test_throws ErrorException diff.forward() # should break
```
matbesancon, line 132: ideally, we should use a more specific exception instead of Error, so defining a custom:
`struct NonConvexException <: Exception end`
and then `throw(NonConvexException())` instead of error(something)

src/diff_model.jl
```diff
@@ -47,15 +47,35 @@ function diff_model(_model::MOI.AbstractOptimizer)
         grads = []
         LHS = create_LHS_matrix(z, λ, Q, G, h, A)
         for param in params
```
matbesancon, line 49: we should add a comment on what this loop on param is doing (what happens for each of the parameter types)

matbesancon:
> @matbesancon am i checking test exception right for non-convex QP?

you meant for testing the part that throws an error right?

AKS1996:
yeah. regarding the error

test/diff_model.jl
AKS1996: ok. i agree. so should we check for convexity ourselves?

matbesancon: in the current implementation, which part is throwing the error?

src/diff_model.jl
AKS1996: added

test/diff_model.jl
AKS1996: its actually `OSQP.jl`. The error is invoked when we call `optimize!` in the `forward` method

matbesancon: then I don't think we should test the error on non-convexity, since it creates a test that works on one solver but not or others.

PR #10, opened 2020-06-09
Include examples in travis

AKS1996:


examples/solve-LP.jl
```diff
@@ -49,35 +48,18 @@ x̄ = MOI.get(model, MOI.VariablePrimal(), x);  # solution
 
 @assert abs(c'x̄ - c'x̂) <= 1e-8   # sanity check
 
-joint_object    = dualize(model)
```
matbesancon, line 52: is it normal that we remove this part?

AKS1996: i was trying to remove `Dualization.jl` as a dependency. As MOI itself provides the dual variables corresponding to a constraint, we don't dualization, as the aim was to verify the KKT conditions

matbesancon: if you don't need it ok, otherwise having Dualization as a dependency is nothing dramatic

PR #17, opened 2020-06-15
Add more tests

AKS1996:
Added a non trivial QP. Reading matrices/gradients from simple txt files.
Gradients were generated by QPTH


test/diff_model.jl
```diff
@@ -419,4 +419,104 @@ end
...
+
+    @test grads[1] ≈ zeros(3,3) atol=ATOL rtol=RTOL
+    @test grads[2] ≈ zeros(3) atol=ATOL rtol=RTOL
+    @test grads[3] ≈ [0.0 0.0 0.0;
+                    0.0 0.0 -5/3;
+                    0.0 0.0 5/3;
+                    0.0 0.0 -10/3;
+                    0.0 0.0 0.0]   atol=ATOL rtol=RTOL
+    @test grads[4] ≈ [0.0; 1/3; -1/3; 2/3; 0.0]   atol=ATOL rtol=RTOL
 end
```
matbesancon, line 522:
```suggestion
end

```

matbesancon:
I think it would be easier to have the txt files in some /data folder, which can be called in the tests

Project.toml
```diff
@@ -4,12 +4,15 @@ authors = ["Akshay Sharma"]
 version = "0.1.0"
 
 [deps]
+Clp = "e2554f3b-3117-50c0-817c-e040a3ddf72d"
```
blegat, line 7: Move these to `[extras]` and `[targets]` as they are only dependencies for the tests

AKS1996: ok. will do it

Issue #12, opened 2020-06-11
Creating a MOI Optimizer layer


be-apt:
referring these projects:
- https://github.com/JuliaOpt/Dualization.jl/blob/master/src/MOI_wrapper.jl
- https://github.com/tomasfmg/ParametricOptInterface.jl/blob/master/src/ParametricOptInterface.jl
- https://github.com/joaquimg/QuadraticToBinary.jl/blob/master/src/moi.jl

todo:
- [x] PR #19 
- [x] #13 
- [x] #14 
- [x] #22

joaquimg:
This is complete since the delivery of the project.

PR #18, opened 2020-06-17
interface changes

AKS1996:
in view of #12


AKS1996:
i didn't cache the matrices. IMO when we add support for conics, `backward!` method will change majorly, while struct and forward method need not

src/MOI_wrapper.jl
```diff
@@ -0,0 +1,229 @@
...
+## Note 
+Currently supports differentiating linear and quadratic programs only.
+"""
+
+const VI = MOI.VariableIndex
+const CI = MOI.ConstraintIndex
+
+
+mutable struct Optimizer <: MOI.AbstractOptimizer
+    optimizer::MOI.Bridges.LazyBridgeOptimizer
```
blegat, line 14: Use `Optimizer{OT<:MOI.ModelLike}` and here do `optimizer::OT` so that it's a concrete type. Otherwise, the performance will be terrible. See https://docs.julialang.org/en/v1/manual/performance-tips/index.html#Avoid-fields-with-abstract-containers-1

src/MOI_wrapper.jl
```diff
@@ -0,0 +1,229 @@
...
+    return vi
+end
+
+
+function MOI.add_variables(model::Optimizer, N::Int)
+    return VI[MOI.add_variable(model) for i in 1:N]
+end
+
+
+function MOI.add_constraint(model::Optimizer, f::F, s::S) where {F<:MOI.AbstractFunction, S<:MOI.AbstractSet}
```
blegat, line 44: You can just do `f::MOI.AbstractFunction` as you don't need `F`

src/MOI_wrapper.jl
```diff
@@ -0,0 +1,229 @@
...
+    return
+end
+
+
+function MOI.set(model::Optimizer, attr::MOI.ObjectiveSense, sense::MOI.OptimizationSense)
+    return MOI.set(model.optimizer, attr, sense)
+end
+
+
+function MOI.get(model::Optimizer, attr::MOI.ObjectiveSense)
```
blegat, line 71: You can maybe just do `MOI.get(model::Optimizer, attr::MOI.AbstractModelAttribute)` instead of having the same method for each model attributes

src/MOI_wrapper.jl
```diff
@@ -0,0 +1,229 @@
...
+
+
+mutable struct Optimizer <: MOI.AbstractOptimizer
+    optimizer::MOI.Bridges.LazyBridgeOptimizer
+    primal_optimal::Array{Float64}  # solution
+    dual_optimal::Array{Float64}  
+    var_idx::Vector{VI}
+    con_idx::Vector{CI}
+
+    function Optimizer(optimizer_constructor::Type)
```
blegat, line 20: Just do `(optimizer_constructor)`, it may not be a `Type`, it could also be a `Function` or a `MOI.OptimizerWithAttributes`, ...

AKS1996: done

src/MOI_wrapper.jl
AKS1996: fixed

src/MOI_wrapper.jl
AKS1996: done

src/MOI_wrapper.jl
AKS1996: fixed

Issue #24, opened 2020-07-04
Differentiate a simple SOCP


AKS1996:
taking [diffcp](https://github.com/cvxgrp/diffcp) as reference
DiffOpt should be able to support most (all ?) MOI constraints

- [x] take projection on cone constraint #25

AKS1996:
seems like a good manually solvable example - https://github.com/jump-dev/MathOptInterface.jl/blob/master/src/Test/contconic.jl#L789

PR #26, opened 2020-07-07
add projections simple test

AKS1996:
in view of #24. missing methods for extracting matrices A,b,c


src/MOI_wrapper.jl
```diff
@@ -129,6 +130,26 @@ function MOI.set(model::Optimizer, attr::MOI.ConstraintFunction, ci::CI{F, S}, f
...
+function MOI.set(model::Optimizer, attr::MOI.ConstraintSet, ci::CI{F, S}, s::S) where {F<:SUPPORTED_VECTOR_FUNCTIONS, S<:SUPPORTED_VECTOR_SETS}
+    return MOI.set(model.optimizer,attr,ci,s)
+end
+
+function MOI.get(model::Optimizer, attr::MOI.ConstraintFunction, ci::CI{F, S}) where {F<:SUPPORTED_VECTOR_FUNCTIONS, S<:SUPPORTED_VECTOR_SETS}
+    return MOI.get(model.optimizer, attr, ci)
+end
+
+function MOI.set(model::Optimizer, attr::MOI.ConstraintFunction, ci::CI{F, S}, f::F) where {F<:SUPPORTED_VECTOR_FUNCTIONS, S<:SUPPORTED_VECTOR_SETS}
+    return MOI.set(model.optimizer,attr,ci,f)
```
matbesancon, line 150:
```suggestion
    return MOI.set(model.optimizer, attr, ci, f)
```

src/MOI_wrapper.jl
```diff
@@ -418,3 +443,63 @@ end
 function MOI.modify(model::Optimizer, ci::CI{F, S}, chg::MOI.AbstractFunctionModification) where {F<:MOI.VectorAffineFunction{Float64}, S <: SUPPORTED_VECTOR_SETS}
     MOI.modify(model.optimizer, ci, chg)
 end
+
+"""
+    Method to differentiate optimal solution `x`, `y`, `s`
+"""
+function backward_conic!(model::Optimizer)
+    if MOI.get(model, MOI.TerminationStatus()) in [MOI.LOCALLY_SOLVED, MOI.OPTIMAL]
+        @assert MOI.get(model.optimizer, MOI.SolverName()) == "SCS"
```
matbesancon, line 452: is this verification necessary?

src/utils.jl
```diff
@@ -133,3 +133,69 @@ function get_problem_data(model::MOI.AbstractOptimizer)
...
+            result[1] = 1.0
+            result[2:length(z)] = x / norm_x
+            result *= (norm_x + t) / 2.0
+            return result
+        end
+    end
+end
+
+"""
+    derivative of the projection of vector `z` on MOI set `cone`
```
matbesancon, line 165: You should specify how is the derivative expressed, is it a matrix with D[i,j] = d proj_i / d  z_i

src/utils.jl
```diff
@@ -133,3 +133,69 @@ function get_problem_data(model::MOI.AbstractOptimizer)
     
     return Q, q, G, h, A, b, nz, var_idx, nineq, ineq_con_idx, neq, eq_con_idx
 end
+
+
+"""
+    projection of vector `z` on MOI set `cone`
+"""
+function π(cone::MOI.AbstractVectorSet, z::Array{Float64})
+    if cone isa MOI.Zeros
```
joaquimg, line 142: Dispatch on cone types instead of doing the conditionals

test/MOI_wrapper.jl
```diff
@@ -532,3 +532,30 @@ end
...
+    MOI.optimize!(model)
+
+    grad_projection = backward_conic!(model)
+
+    @test grad_projection ≈ [1.          0.          0.          0.          0.        ;
+                   0.          1.          0.          0.          0.        ;
+                   0.          0.          0.5         0.35355328 -0.3535535 ;
+                   0.          0.          0.35355328  0.54289328  0.04289326;
+                   0.          0.         -0.3535535   0.04289326  0.54289323] atol=ATOL rtol=RTOL
+end
```
joaquimg, line 561: you got these results from python?

src/utils.jl
```diff
@@ -133,3 +133,69 @@ function get_problem_data(model::MOI.AbstractOptimizer)
...
+        return zeros(Float64, size(z))
+    elseif cone isa MOI.Nonnegatives
```
joaquimg, line 144: you can also add Nonpositives, the maths should be very simple

test/MOI_wrapper.jl
AKS1996: yes. can add a simple `print` command in https://github.com/cvxgrp/diffcp

src/utils.jl
AKS1996: ok. will do it

src/MOI_wrapper.jl
AKS1996: actually not :sweat_smile: i didn't knew if any other solver also expresses the matrices A,b,c as SCS.jl does

src/utils.jl
```diff
@@ -133,3 +133,69 @@ function get_problem_data(model::MOI.AbstractOptimizer)
...
+            result *= (norm_x + t) / 2.0
+            return result
+        end
+    end
+end
+
+"""
+    derivative of the projection of vector `z` on MOI set `cone`
+"""
+function Dπ(cone::MOI.AbstractVectorSet, z::Array{Float64})
```
matbesancon, line 167: The function is type-unstable, with sometimes a matrix, sometimes a vector being returned

blegat: Is it an issue if it's only the union of two types ?

src/utils.jl
AKS1996: added

src/utils.jl
AKS1996: actually for `BlockDiagonals` to work, i need to turn vectors returned by this method into matrices

AKS1996: will change it here only. that way the function will be type stable too

test/MOI_wrapper.jl
joaquimg: add that as a comment here

src/utils.jl
joaquimg: for instance:
```julia
function π(cone::MOI.Zeros, z::Array{Float64})
...
end
function π(cone::MOI.Nonnegatives, z::Array{Float64})
...
end
```

AKS1996: done

AKS1996:
don't know why travis is not showing up. but [build works fine](https://travis-ci.org/github/AKS1996/DiffOpt.jl)

test/MOI_wrapper.jl
joaquimg: @AKS1996 just add this and goto go for me!

AKS1996: added

Issue #27, opened 2020-07-09
Add MOI standard tests


joaquimg:
Do like this package here: https://github.com/joaquimg/QuadraticToBinary.jl/blob/master/test/runtests.jl

This will make the API more robust and closer to what JuMP needs.

AKS1996:
almost all tests added - left with "From MOI ncqcp", "From MOI qcp", and integer tests

matbesancon:
@joaquimg should we close this one now that MOI tests are there

PR #28, opened 2020-07-11
add basic tests

AKS1996:
in reference to #27


src/MOI_wrapper.jl
```diff
@@ -396,12 +391,19 @@ function _constraint_contains(model::Optimizer, v::VI, ci::CI{F,S}) where {F <:
...
         return v == func.variable
     elseif func isa MOI.ScalarAffineFunction{Float64}
-        for term in func.terms
-            if term.variable_index == v
-                return true
-            end
-        end
-        return false
+        indices = [term.variable_index for term in func.terms]
+        return v in indices
```
matbesancon, line 395:
```suggestion
        return any(term -> v == term.variable, func.terms)
```
This will be equivalent and avoids allocating an additional `indices` array

matbesancon: Same thing below in `_constraint_contains`

src/MOI_wrapper.jl
```diff
@@ -396,12 +391,19 @@ function _constraint_contains(model::Optimizer, v::VI, ci::CI{F,S}) where {F <:
...
+    end
+    return false   # default
+end
+
+function _constraint_contains(model::Optimizer, v::VI, ci::CI{F,S}) where {F <: SUPPORTED_VECTOR_FUNCTIONS, S <: SUPPORTED_VECTOR_SETS}
```
matbesancon, line 400: Rewrite `_constraint_contains` as to use dispatch on the `F` type:
```julia
 function _constraint_contains(model::Optimizer, v::VI, ci::CI{F,S}) where {F <: MOI.VectorOfVariables, S <: SUPPORTED_VECTOR_SETS}
...
 function _constraint_contains(model::Optimizer, v::VI, ci::CI{F,S}) where {F <: MOI.VectorAffineFunction, S <: SUPPORTED_VECTOR_SETS}
...
```

AKS1996: will it improve efficiency?

AKS1996: coz `SUPPORTED_VECTOR_FUNCTIONS` is nothing more than a Union of both

src/MOI_wrapper.jl
AKS1996: fixed

AKS1996:
will need to fix `update_dimension_nonnegative_variables` test though. i think i'm deleting constraints related to constrained variables (doesn't happen in model created by `MOI.instantiate`)

src/MOI_wrapper.jl
joaquimg: Yes, make all the difference for efficiency.
This kind of statement is terrible: `func isa MOI.VectorOfVariables` for performant julia.
Doing that as @matbesancon suggested is the way to go.

AKS1996: ok fixed it

AKS1996: will keep it in mind for future

src/MOI_wrapper.jl
```diff
@@ -31,6 +31,7 @@ const SUPPORTED_SCALAR_FUNCTIONS = Union{
 }
 
 const SUPPORTED_VECTOR_FUNCTIONS = Union{
+    MOI.VectorOfVariables,
     MOI.VectorAffineFunction{Float64}
```
matbesancon, line 35:
```suggestion
    MOI.VectorAffineFunction{Float64},
```

src/MOI_wrapper.jl
```diff
@@ -345,33 +346,27 @@ function MOI.supports(model::Optimizer, ::MOI.VariablePrimalStart,
...
-end
-
 function MOI.delete(model::Optimizer, ci::CI{F,S}) where {F <: SUPPORTED_SCALAR_FUNCTIONS, S <: SUPPORTED_SCALAR_SETS}
     filter!(e -> e≠ci, model.con_idx)
     MOI.delete(model.optimizer, ci) 
 end
 
-function MOI.delete(model::Optimizer, ci::CI{F,S}) where {F <: MOI.VectorAffineFunction{Float64}, S <: SUPPORTED_VECTOR_SETS}
+function MOI.delete(model::Optimizer, ci::CI{F,S}) where {F <: SUPPORTED_VECTOR_FUNCTIONS, S <: SUPPORTED_VECTOR_SETS}
     filter!(e -> e≠ci, model.con_idx)
```
matbesancon, line 360:
```suggestion
    filter!(≠(ci), model.con_idx)
```

`≠(ci)` creates an anonymous function-like object that does the same thing as x -> x != y

AKS1996: thanks! i'll read up more on it

src/MOI_wrapper.jl
AKS1996: does this also help?

AKS1996: i guess i've to read a lot about Julia yet :sweat_smile:

src/MOI_wrapper.jl
AKS1996: hey. i wasn't able to execute this. even a simple script like
```julia
a = [1,2,3]
x = 2
filter(≠(x), a)
```
with 
```
a = [1,2,3]
x = 2
filter!(≠(x), a)
a = [1,2,3]
x = 2
filter!(≠(x), a)
MethodError: no method matching !=(::Int64)
Closest candidates are:
  !=(::Any, !Matched::Any) at operators.jl:185
  !=(!Matched::Float64, !Matched::Float64) at float.jl:450
  !=(!Matched::Float32, !Matched::Float32) at float.jl:449
  ...
```

matbesancon:
for the !=(3) thing, indeed this is a julia >= 1.2 feature so not previously supported

AKS1996:
working finally :sweat_smile:  updated to latest version for now

matbesancon:
I'm not sure we want to remove support for julia 1.0, I would be for it but JuMP is still currently supporting 1.0 while waiting for the next LTS

AKS1996:
Ok. I'll try to add support for Julia 1.0 too. Will just need to edit that delete method

matbesancon:
not sure it's necessary, @joaquimg @blegat what do you think?

blegat:
I think it's best to keep support for Julia v1.0

AKS1996:
ok. right now i've included support for both 1.0 and 1.4

AKS1996:
if there's the tests included are ok, can i merge this branch. a PR for SDP coming tonight :smile:

Project.toml
```diff
@@ -12,7 +12,7 @@ MathOptInterface = "b8f27783-ece8-5eb3-8dc8-9495eed66fee"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 
 [compat]
-julia = "1.0"
+julia = "1.4"
```
blegat, line 15: This should be "1"

src/MOI_wrapper.jl
```diff
@@ -352,12 +352,20 @@ function MOI.get(model::Optimizer, attr::MOI.VariablePrimal, vi::VI)
 end
 
 function MOI.delete(model::Optimizer, ci::CI{F,S}) where {F <: SUPPORTED_SCALAR_FUNCTIONS, S <: SUPPORTED_SCALAR_SETS}
-    filter!(e -> e≠ci, model.con_idx)
+    if VERSION.minor >= 2
```
blegat, line 355: Use `@static if VERSION >= v"1.2"`

joaquimg:
I think its good to go after fixing Benoît's suggestions

Project.toml
AKS1996: added

src/MOI_wrapper.jl
AKS1996: added

PR #30, opened 2020-07-20
Support PSD constraints

AKS1996:
still some fixes to go

### DIFF
--- a/src/MOI_wrapper.jl
+++ b/src/MOI_wrapper.jl
@@ -40,7 +40,7 @@ const SUPPORTED_VECTOR_SETS = Union{
     MOI.Nonpositives,
     MOI.Nonnegatives,
     MOI.SecondOrderCone,
-    MOI.PositiveSemidefiniteConeTriangle
+    MOI.PositiveSemidefiniteConeTriangle,
 }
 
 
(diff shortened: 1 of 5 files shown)


src/MOI_wrapper.jl
```diff
@@ -39,7 +39,8 @@ const SUPPORTED_VECTOR_SETS = Union{
     MOI.Zeros, 
     MOI.Nonpositives,
     MOI.Nonnegatives,
-    MOI.SecondOrderCone
+    MOI.SecondOrderCone,
+    MOI.PositiveSemidefiniteConeTriangle
```
matbesancon, line 43:
```suggestion
    MOI.PositiveSemidefiniteConeTriangle,
```

src/MOI_wrapper.jl
```diff
@@ -457,45 +457,27 @@ end
...
-            c = [1.0; 0.0; 0.0]
+        for con in model.con_idx
+            func = MOI.get(model.optimizer, MOI.ConstraintFunction(), con)
+            set = MOI.get(model.optimizer, MOI.ConstraintSet(), con)
+            MOIU.load_constraint(model.optimizer.model.optimizer, CI{typeof(func), typeof(set)}(0), func, set)
         end
+        
+        # now SCS data shud be allocated
+        A = sparse(
+            model.optimizer.model.optimizer.data.I,
```
matbesancon, line 471: `model.optimizer.model.optimizer.data.I,`

these lines look specific to one unique solver right?

src/MOI_wrapper.jl
```diff
@@ -457,45 +457,27 @@ end
...
-                0.0   0.0  -1.0;
-                -1.0  0.0   0.0;
-                0.0  -1.0   0.0;
-            ]
-            b = [1.0; -0.707107; 0.0; 0.0; 0.0]
-            c = [1.0; 0.0; 0.0]
+        for con in model.con_idx
+            func = MOI.get(model.optimizer, MOI.ConstraintFunction(), con)
+            set = MOI.get(model.optimizer, MOI.ConstraintSet(), con)
+            MOIU.load_constraint(model.optimizer.model.optimizer, CI{typeof(func), typeof(set)}(0), func, set)
```
matbesancon, line 466: why 0 here?

src/MOI_wrapper.jl
AKS1996: yes. extracting the matrices for SCS.jl was easiest

AKS1996: actually this is  the same code used in SCS.jl's optimize method https://github.com/jump-dev/SCS.jl/blob/master/src/MOI_wrapper.jl#L420

src/MOI_wrapper.jl
AKS1996: added

src/MOI_wrapper.jl
AKS1996: I'm not sure why, but when we run MOI.optimize! on a SCS.Optimizer model, this method load_constraint is called with `ConstraintIndex` for all constraints set as 0

AKS1996: had a rough time understanding why it is so. but this "hack" works to populate matrices A,b,c correctly (matched against diffcp)

src/MOI_wrapper.jl
```diff
@@ -152,7 +153,7 @@ end
 
 
 function MOI.optimize!(model::Optimizer)
-    MOI.optimize!(model.optimizer)
+    solution = MOI.optimize!(model.optimizer)
```
blegat, line 156: `optimize!` returns `nothing`

src/MOI_wrapper.jl
blegat: I wouldn't use the allocate-load API, just use `add_constraint`, you can check whether the solver needs a caching optimizer with `supports_default_copy_to`

blegat: That way, you will use a cache for SCS but not for Mosek for instance

blegat: By the way, the constraint index should be the one returned by load_constraint

src/MOI_wrapper.jl
AKS1996: for SCS.jl, it returns `MOISolution`. can query it to get slack and dual solution easily

AKS1996: added this only for SCS.jl. indeed other solvers dont return anything

PR #37, opened 2020-08-20
add MatOI as a dependency

AKS1996:
will merge once https://github.com/jump-dev/MatrixOptInterface.jl/pull/7 is complete


AKS1996:
https://github.com/jump-dev/MatrixOptInterface.jl/pull/7 is merged, but the code still contains some SCS specific preprocessing. Will update the PR after removing it

AKS1996:
major revisions. update some tests, gradients are much more intuitive now
will work once https://github.com/jump-dev/MatrixOptInterface.jl/pull/10 is merged

matbesancon:
@blegat @AKS1996 will this be fixing #47 ? If yes we can move this forward, since tests are currently broken?

PR #40, opened 2020-08-26
time/memory benchmarking

AKS1996:
solves https://github.com/AKS1996/DiffOpt.jl/issues/11
in need some guidance. i'm total noob at benchmarking. but i was thinking to improve both sparsity and time in same PR

### DIFF
--- a/src/MOI_wrapper.jl
+++ b/src/MOI_wrapper.jl
@@ -43,6 +43,8 @@ const SUPPORTED_VECTOR_SETS = Union{
     MOI.PositiveSemidefiniteConeTriangle,
 }
 
+const SUPPORTED_TERMINATION_STATUS = [MOI.LOCALLY_SOLVED, MOI.OPTIMAL, MOI.ALMOST_OPTIMAL]
+
 """
     diff_optimizer(optimizer_constructor)::Optimizer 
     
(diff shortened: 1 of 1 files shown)


AKS1996:
using this script for julia benchmarking (in need of suggestions :Ddiffer

src/MOI_wrapper.jl
```diff
@@ -43,6 +43,8 @@ const SUPPORTED_VECTOR_SETS = Union{
     MOI.PositiveSemidefiniteConeTriangle,
 }
 
+const SUPPORTED_TERMINATION_STATUS = [MOI.LOCALLY_SOLVED, MOI.OPTIMAL, MOI.ALMOST_OPTIMAL]
```
matbesancon, line 46: When do we have cases with `MOI.LOCALLY_SOLVED`?

AKS1996: yep. i remember differentiating a convex QP in some solver gave this result. let me fetch the history

matbesancon:
> `but i was thinking to improve both sparsity and time in same PR`

I would suggest sparsity in a different PR not to get things confused

AKS1996:
> I would suggest sparsity in a different PR not to get things confused

noted will create a new PR

matbesancon:
Your Python benchmark seems to profile memory usage more than time?

AKS1996:
as i said, noob at profiling :sweat_smile:
but i ran them separately (one at a time) so there should'nt be any overlap
although i'm not sure these results are from single or multiple iterations

matbesancon:
https://docs.python.org/3/library/timeit.html
this module should do what you want

AKS1996:
PR doesn't contains any code/contribution for the project 
closing it

PR #41, opened 2020-08-26
Support sparsity

AKS1996:
in light of https://github.com/AKS1996/DiffOpt.jl/issues/31

### DIFF
--- a/src/MOI_wrapper.jl
+++ b/src/MOI_wrapper.jl
@@ -275,7 +275,7 @@ function backward!(model::Optimizer, params::Array{String}, dl_dz::Array{Float64
     LHS = create_LHS_matrix(z, λ, Q, G, h, A)
     RHS = sparse([dl_dz'; zeros(neq+nineq,1)])
 
-    partial_grads = -(LHS \ RHS)
+    partial_grads = -lsqr(LHS, RHS)   # error rate for lsqr should be small comparison to (LHS \ RHS)
 
     dz = partial_grads[1:nz]
     if nineq > 0
--- a/src/MOI_wrapper.jl
+++ b/src/MOI_wrapper.jl
@@ -43,7 +43,7 @@ const SUPPORTED_VECTOR_SETS = Union{
     MOI.PositiveSemidefiniteConeTriangle,
 }
 
-const SUPPORTED_TERMINATION_STATUS = [MOI.LOCALLY_SOLVED, MOI.OPTIMAL, MOI.ALMOST_OPTIMAL]
+const SUPPORTED_TERMINATION_STATUS = [MOI.LOCALLY_SOLVED, MOI.OPTIMAL, MOI.ALMOST_OPTIMAL, MOI.ALMOST_LOCALLY_SOLVED]
 
 """
     diff_optimizer(optimizer_constructor)::Optimizer 
@@ -273,9 +273,9 @@ function backward!(model::Optimizer, params::Array{String}, dl_dz::Array{Float64
 
     grads = []
     LHS = create_LHS_matrix(z, λ, Q, G, h, A)
-    RHS = sparse([dl_dz'; zeros(neq+nineq,1)])
+    RHS = [dl_dz'; zeros(neq+nineq)]
 
-    partial_grads = -lsqr(LHS, RHS)   # error rate for lsqr should be small comparison to (LHS \ RHS)
+    partial_grads = -LHS \ RHS
 
     dz = partial_grads[1:nz]
     if nineq > 0
@@ -522,7 +522,7 @@ end
 Find projection of vectors in `v` on product of `cones`.
 For more info, refer https://github.com/matbesancon/MathOptSetDistances.jl
 """
-π(cones, v) = sparse(MOSD.projection_on_set(MOSD.DefaultDistance(), v, cones))
+π(cones, v) = MOSD.projection_on_set(MOSD.DefaultDistance(), v, cones)
 
 
 """
@@ -531,7 +531,7 @@ For more info, refer https://github.com/matbesancon/MathOptSetDistances.jl
 Find gradient of projection of vectors in `v` on product of `cones`.
 For more info, refer https://github.com/matbesancon/MathOptSetDistances.jl
 """
-Dπ(cones, v) = sparse(MOSD.projection_gradient_on_set(MOSD.DefaultDistance(), v, cones))
+Dπ(cones, v) = MOSD.projection_gradient_on_set(MOSD.DefaultDistance(), v, cones)
 
 """
     backward_conic!(model::Optimizer, dA::Array{Float64,2}, db::Array{Float64}, dc::Array{Float64})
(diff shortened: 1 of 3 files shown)


AKS1996:
tested differentiating simple LP - https://github.com/AKS1996/jump-gsoc-2020/blob/master/DiffOpt_time_benchmarking.ipynb

just adding support as basic SparseArrays reduced memory usage to a quarter 

```bash
BenchmarkTools.Trial: 
  memory estimate:  38.50 MiB
  allocs estimate:  229521
  --------------
  minimum time:     78.738 ms (2.75% GC)
  median time:      83.476 ms (5.46% GC)
  mean time:        84.282 ms (4.85% GC)
  maximum time:     94.965 ms (2.16% GC)
  --------------
  samples:          60
  evals/sample:     1
```

src/MOI_wrapper.jl
```diff
@@ -596,31 +598,35 @@ function backward_conic!(model::Optimizer, dA::Array{Float64,2}, db::Array{Float
...
-        M = ((Q - Matrix{Float64}(I, size(Q))) * BlockDiagonal([Matrix{Float64}(I, length(u), length(u)), Dπv, Matrix{Float64}(I,1,1)])) + Matrix{Float64}(I, size(Q))
+        M = ((Q - sparse(1.0I, size(Q))) * blockdiag(sparse(1.0I, length(u), length(u)), Dπv, sparse(1.0I, 1, 1))) + sparse(1.0I, size(Q))
 
         # find projections on dual of the cones
-        πz = vcat([u, π([MOI.dual_set(cone) for cone in cones], v), max.(w,0.0)]...)
+        πz = sparse(vcat([u, π([MOI.dual_set(cone) for cone in cones], v), max.(w,0.0)]...))
 
-        dQ = [
+        dQ = sparse([
             zeros(n,n)    dA'          dc;
```
joaquimg, line 620: use spzeros instead of zeros

src/MOI_wrapper.jl
```diff
@@ -596,31 +598,35 @@ function backward_conic!(model::Optimizer, dA::Array{Float64,2}, db::Array{Float
         z = (x, y - s, [1.0])
         u, v, w = z
 
-        Q = [
+        # reshaping vectors to matrix
+        b = b[:, :]
+        c = c[:, :]
+
+        Q = sparse([
             zeros(n,n)   A'           c;
```
joaquimg, line 606: spzeros

src/MOI_wrapper.jl
```diff
@@ -596,31 +598,35 @@ function backward_conic!(model::Optimizer, dA::Array{Float64,2}, db::Array{Float
         z = (x, y - s, [1.0])
         u, v, w = z
 
-        Q = [
+        # reshaping vectors to matrix
+        b = b[:, :]
+        c = c[:, :]
```
joaquimg, line 603: why do we need these two lines?

src/MOI_wrapper.jl
```diff
@@ -43,6 +43,8 @@ const SUPPORTED_VECTOR_SETS = Union{
     MOI.PositiveSemidefiniteConeTriangle,
 }
 
+const SUPPORTED_TERMINATION_STATUS = [MOI.LOCALLY_SOLVED, MOI.OPTIMAL, MOI.ALMOST_OPTIMAL]
```
joaquimg, line 46: there is also, ALMOST_LOCALLY_SOLVED

src/MOI_wrapper.jl
```diff
@@ -271,9 +273,9 @@ function backward!(model::Optimizer, params::Array{String}, dl_dz::Array{Float64
 
     grads = []
     LHS = create_LHS_matrix(z, λ, Q, G, h, A)
-    RHS = [dl_dz'; zeros(neq+nineq,1)]
+    RHS = sparse([dl_dz'; zeros(neq+nineq,1)])
```
joaquimg, line 276: shouldnt it be a vector rather than a one column matrix?

src/MOI_wrapper.jl
```diff
@@ -529,7 +531,7 @@ For more info, refer https://github.com/matbesancon/MathOptSetDistances.jl
 Find gradient of projection of vectors in `v` on product of `cones`.
 For more info, refer https://github.com/matbesancon/MathOptSetDistances.jl
 """
-Dπ(cones, v) = MOSD.projection_gradient_on_set(MOSD.DefaultDistance(), v, cones)
+Dπ(cones, v) = sparse(MOSD.projection_gradient_on_set(MOSD.DefaultDistance(), v, cones))
```
joaquimg, line 534: not sure if we force sparsity here or where Dπ is used

src/MOI_wrapper.jl
```diff
@@ -596,31 +598,35 @@ function backward_conic!(model::Optimizer, dA::Array{Float64,2}, db::Array{Float
...
             zeros(n,n)    dA'          dc;
             -dA           zeros(m,m)   db;
             -dc'         -db'          zeros(1,1)
-        ]
+        ])
 
         RHS = dQ * πz
         if norm(RHS) <= 1e-4
-            dz = zeros(Float64, size(RHS))
+            dz = sparse(0.0I, size(RHS[:, :]))
```
joaquimg, line 627: do we need this [:,:] ?

src/MOI_wrapper.jl
AKS1996: reindex/convert SparseVector to SparseMatrix. turns out SparseVector's give error on multiplication

src/MOI_wrapper.jl
AKS1996: ok. adding

src/MOI_wrapper.jl
AKS1996: yep. i just wanted to test how much we can improve on benchmarking. projections aren't that sparse, and ideally MOSD should return sparse values

src/MOI_wrapper.jl
AKS1996: yep, again RHS turns out to be a SparseVector

AKS1996:
i know i've added `[:, :]` too may times. i'll try to see where is it redundant

src/MOI_wrapper.jl
matbesancon: forcing sparsity of the type if the projection itself isn't sparse will result in poor performance

src/MOI_wrapper.jl
AKS1996: changed

src/MOI_wrapper.jl
AKS1996: changed

src/MOI_wrapper.jl
AKS1996: yep, changed it

src/MOI_wrapper.jl
AKS1996: ok, removing it from here

src/MOI_wrapper.jl
AKS1996: removed it

src/MOI_wrapper.jl
```diff
@@ -596,31 +598,35 @@ function backward_conic!(model::Optimizer, dA::Array{Float64,2}, db::Array{Float
...
-        dQ = [
-            zeros(n,n)    dA'          dc;
-            -dA           zeros(m,m)   db;
-            -dc'         -db'          zeros(1,1)
-        ]
+        dQ = sparse([
+            spzeros(n,n)    dA'          dc;
+            -dA           spzeros(m,m)   db;
+            -dc'         -db'          spzeros(1,1)
+        ])
```
matbesancon, line 623: this part is doing much conversion work.
```
[
    a  b  c
    a  b  c
]
```

creates a dense matrix out of the sparse arrays, which is then re-converted to sparse.
It would be better to create an empty dQ with spzeros(n+m+1, n+m+1) and then fill the non-diagonal blocks
with
dQ[n+1:n+1+m,1:n] .= -dA
etc.

Same thing above for the creation of Q

AKS1996: oh. thanks for pointing out. never realized this redundant sparse-dense-sparse conversion

matbesancon: I think the spzeros(1,1) can be replaced by just `0`

matbesancon (approved):

src/MOI_wrapper.jl
AKS1996: fixed

matbesancon:
@AKS1996 why close this one?

AKS1996:
there's a lot of code that's outdated (removed coz of SCS or moved to MatOI). it'll be suitable to add sparsity support to MatOI too. Once DiffOpt builds w/o errors, I'll do memory benchmarking and open separate PRs to support sparsity

AKS1996:
if you want, we can open this one too

PR #45, opened 2020-12-03
fix URLs, github actions

matbesancon:
Fix various small URLs to `jump-dev` and replace travis and appveyor with github actions


matbesancon:
The tests seem to use an incorrect signature of MOI.optimize!
```
    # model = DiffOpt.Optimizer(...)

    sol = MOI.optimize!(model)

    x = sol.primal
    s = sol.slack
    y = sol.dual

```

in MOI_wrapper.jl.

Ping @AKS1996

matbesancon:
The primal and dual access parts can be fixed quickly, but the slack requires some additional computation

matbesancon:
Fixing CI does not make sense while the tests are all still breaking, so tackling the immediate issues
#46 #47

matbesancon:
I am wondering if we should merge this one first and then fix the later issues

matbesancon:
OK all tests are passing except for the ones linked to SCS not supporting a given constraint type, i.e. linked to
https://github.com/jump-dev/DiffOpt.jl/pull/37
I would suggest merging this and getting #37 merged soon. The other way would be to comment out all conic tests (not ideal) or merge #37 into this

blegat (approved):
Let's merge

matbesancon:
I made the tests pass, I'll just update for GH actions to add MatrixOptI

matbesancon:
Not sure about the GH syntax, we'll see

matbesancon:
So good news is some tests pass and the actions are finally set up.
Bad news is there are still some failing tests

matbesancon:
I suspect the test problems not to be strongly convex -> multiple possible solutions found by different solvers?

matbesancon:
There seems to be a windows-only escaping issue:
https://github.com/jump-dev/DiffOpt.jl/pull/45/checks?check_run_id=1517097034
when Pkg.dev-ing MatOI. @odow have you encountered something similar?

matbesancon:
I get rather large numerical errors on a slack test:
```
s - expected_s:
 1.1734166677389504e-15
 0.13905758625384124
 4.715214951276747e-16
 0.1775338024275932
 6.69285964927494e-17
 5.279372035384908e-16
 3.843926470251321e-17
 0.2625892873538369
 0.04306063761061951
 0.1012750010480592
 0.0
```

matbesancon:
I think we're close to something working for docs too, once this is the case this should be merged, I still don't know how to make windows work with URL properly without breaking on Linux. And the 32-bits issue is related to an upstream package

matbesancon:
alright good to merge

matbesancon:
Ok windows passing too

matbesancon:
replaces #37

matbesancon:
bump

src/utils.jl
```diff
@@ -88,54 +86,78 @@ function get_problem_data(model::MOI.AbstractOptimizer)
...
+
+        q = MOI.coefficient.(objective_function.affine_terms)
+    end
+
+    return Q, q, G, h, A, b, nz, var_list, nineq, ineq_con_idx, neq, eq_con_idx
+end
+
+# might slow down computation
+# need to find a faster way
+function CSRToCSC(B::MatOI.SparseMatrixCSRtoCSC{Int64})
```
blegat, line 154: In fact, it's easier to get the conic form for CSC A, we do it in CSR in MatOI as this is what SCS needs but we can add a mutable CSC matrix in MatOI too quite easily so that DiffOpt can use that but let's leave that for another PR

blegat (approved):

src/utils.jl
matbesancon: maybe document it in an issue?

matbesancon:
See https://github.com/oxfordcontrol/OSQP.jl/issues/68 for removing 32 bits for now

PR #56, opened 2021-01-17
Minor formatting

matbesancon:
Minor aesthetic things


matbesancon:
using the fixed syntax `!=(x)` is not worth the trouble of keeping two branches of the same code for julia 1.0 and the newer ones

PR #59, opened 2021-01-22
matrix-less dQ

matbesancon:
In backward_conic, `dQ` is materialized only for a matrix-vector product, it is eliminated from the pass.

There should also be a way to avoid materializing `Q`, and ultimately reasoning on the sparse construction of the problem and not its explicit matrix form

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # DiffOpt.jl
 
-[![Stable](https://img.shields.io/badge/docs-stable-blue.svg)](https://jump-dev.org/DiffOpt.jl/stable)
-[![Dev](https://img.shields.io/badge/docs-dev-blue.svg)](https://jump-dev.org/DiffOpt.jl/dev)
+[![Stable](https://img.shields.io/badge/docs-stable-blue.svg)](https://jump.dev/DiffOpt.jl/stable)
+[![Dev](https://img.shields.io/badge/docs-dev-blue.svg)](https://jump.dev/DiffOpt.jl/dev)
 [![Build Status](https://github.com/jump-dev/DiffOpt.jl/workflows/CI/badge.svg)](https://github.com/jump-dev/DiffOpt.jl/actions)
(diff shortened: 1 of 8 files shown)


matbesancon:
Good to merge

README.md
```diff
@@ -1,16 +1,17 @@
...
 [![Coverage](https://codecov.io/gh/jump-dev/DiffOpt.jl/branch/master/graph/badge.svg)](https://codecov.io/gh/jump-dev/DiffOpt.jl)
 
 Differentiating convex optimization program (`JuMP.jl` or `MathOptInterface.jl` models) with respect to program parameters. Currently supports LPs, QPs.
 
 ## Installation
 DiffOpt can be installed through the Julia package manager:
+
 ```
-(v1.3) pkg> add https://github.com/jump-dev/DiffOpt.jl
+(v1.3) pkg> add DiffOpt
```
blegat, line 14: Why is that ? It is not released yet

matbesancon: nope you're right, out of habit. Fixed

matbesancon:
if someone wants to have a final look before merging this

src/MOI_wrapper.jl
```diff
@@ -527,40 +526,49 @@ function backward_conic!(model::Optimizer, dA::Array{Float64,2}, db::Array{Float
...
-        dz = zeros(Float64, size(RHS))
+    dz = if norm(RHS) <= 1e-4
+        RHS .= 0
     else
-        dz = lsqr(M, RHS)
+        lsqr(M, RHS)
     end
 
-    du, dv, dw = dz[1:n], dz[n+1:n+m], dz[n+m+1]
+    @inbounds (du, dv, dw) = (dz[1:n], dz[n+1:n+m], dz[n+m+1])
```
blegat, line 568: This `@inbounds` may trigger more segfault (when we change this function and forget something) than have any visible effect on perf no ? This isn't a critical part of the code.

matbesancon: yes right, normally we control the size of dz deduced from the RHS we form but the bound check might be more future-proof

matbesancon: updated

Issue #61, opened 2021-01-27
Cache gradients


matbesancon:
Each backward solve currently solves the full linear system, we could instead cache the already-computed gradients the Optimizer, invalidate the cache when the problem is modified, and compute in backward lazily

Issue #65, opened 2021-01-31
Unused variables create singular linear solve exceptions


matbesancon:
MWE:

```julia
    Q = [
        4.0 1.0
        1.0 2.0
    ]
    q = [1.0, 1.0]
    G = [1.0 1.0]
    h = [-1.0]

    model = diff_optimizer(OSQP.Optimizer)
    MOI.set(model, MOI.Silent(), true)
    x = MOI.add_variables(model, 2)

    # define objective
    quad_terms = MOI.ScalarQuadraticTerm{Float64}[]
    for i in 1:2
        for j in i:2 # indexes (i,j), (j,i) will be mirrored. specify only one kind
            push!(
                quad_terms,
                MOI.ScalarQuadraticTerm(Q[i,j], x[i], x[j])
            )
        end
    end
    objective_function = MOI.ScalarQuadraticFunction(
                            MOI.ScalarAffineTerm.(q, x),
                            quad_terms,
                            0.0,
                        )
    MOI.set(model, MOI.ObjectiveFunction{MOI.ScalarQuadraticFunction{Float64}}(), objective_function)
    MOI.set(model, MOI.ObjectiveSense(), MOI.MIN_SENSE)

    # add constraint
    MOI.add_constraint(
        model,
        MOI.ScalarAffineFunction(MOI.ScalarAffineTerm.(G[1, :], x), 0.0),
        MOI.LessThan(h[1]),
    )
    MOI.optimize!(model)
    
    @test model.primal_optimal ≈ [-0.25; -0.75] atol=ATOL rtol=RTOL
    @test isempty(model.gradient_cache)
    grad_wrt_h = backward!(model, ["h"], ones(2))[1]
    @test grad_wrt_h ≈ [1.0] atol=2ATOL rtol=RTOL
    @test !isempty(model.gradient_cache)

    # adding two variables invalidates the cache
    y = MOI.add_variables(model, 2)
    for yi in y
        MOI.delete(model, yi)
    end
    @test isempty(model.gradient_cache)
    MOI.optimize!(model)
    grad_wrt_h = backward!(model, ["h"], ones(2))[1]
    @test grad_wrt_h ≈ [1.0] atol=2ATOL rtol=RTOL
    @test !isempty(model.gradient_cache)
```

Having new variables `y` that don't get used in constraints makes the linear solve fail:

```julia
  SingularException(0)
  Stacktrace:
[...]

PR #66, opened 2021-01-31
add gradient cache in Optimizer

matbesancon:
Fixes https://github.com/jump-dev/DiffOpt.jl/issues/61

still failing because of #65


matbesancon:
Closes #68

matbesancon:
OSQP appears to have more inaccuracy / variations on the final result, that's why I switched to Ipopt

blegat:
> OSQP appears to have more inaccuracy / variations on the final result, that's why I switched to Ipopt

Yes, when I use it for the tests, I usually increase the accuracy, e.g.
```julia
"eps_abs" => 1e-8,
"eps_rel" => 1e-8,
```

src/MOI_wrapper.jl
```diff
@@ -255,31 +266,44 @@ Note that this method *does not returns* the actual jacobians.
...
     dz = partial_grads[1:nz]
+
     if nineq > 0
         dλ = partial_grads[nz+1:nz+nineq]
     end
     if neq > 0
         dν = partial_grads[nz+nineq+1:nz+nineq+neq]
     end
 
+    grads = []
```
blegat, line 306: Can't we have at least `Union{Vector{Float64}, Matrix{Float64}}[]` ? It's not only for perf, it also helps readability.

matbesancon: This is just a line that was moved from another place in the function, I plan to return gradients in a cleaner way in another one

src/MOI_wrapper.jl
```diff
@@ -73,6 +73,7 @@ mutable struct Optimizer{OT <: MOI.ModelLike} <: MOI.AbstractOptimizer
     dual_optimal::Vector{Union{Vector{Float64}, Float64}}  # refer - https://github.com/jump-dev/DiffOpt.jl/issues/21#issuecomment-651130485
     var_idx::Vector{VI}
     con_idx::Vector{CI}
+    gradient_cache::NamedTuple
```
blegat, line 76: Why not using `Union{Nothing, NamedTuple{...}}` or even `Union{Nothing, Cache}` with an actual struct ? Currently, it's type unstable

matbesancon: This will need at least `Union{Nothing, NamedTuple{T1}, NamedTuple{T2}}` for the cache, since we have possible cache types depending on conic or not

blegat: I'd still prefer `struct Cache1 end; struct Cache2 end; Union{Nothing, Cache1, Cache2}`, otherwise it's not so easy to follow what's going on.
Moreover, it seems to me that `NamedTuple` are used when you don't know a priori what is the struct so it's like creating new structs on-the-fly why here we know very well the struct at coding time so we should use structs as it's more readable.

matbesancon:
bump here

blegat (approved):

blegat (approved):

PR #81, opened 2021-03-15
[WIP] correct forw and back modes

joaquimg:


src/conic_diff.jl
```diff
@@ -0,0 +1,75 @@
...
+        M = M,
+        vp = vp,
+        Dπv = Dπv,
+        xys = (x, y, s),
+        A = A,
+        b = b,
+        c = c,
+    )
+    return nothing
+end
```
matbesancon, line 75:
```suggestion
end

```

matbesancon:
`src/quadratic_diff.jl` has been created empty it seems

matbesancon:
Also this PR should update the documentation

joaquimg:
Thanks for the review!
I will keep updating this week.
I am doing a bit too much in pieces

joaquimg:
Next steps include:
 - query grad sparsely (MOI like)
 - send seed perturbations sparsely (MOI like)
 - docs
 - tests

src/MOI_wrapper.jl
```diff
@@ -92,25 +91,62 @@ Base.@kwdef struct ConicCache
...
     QPCache,
     ConicCache,
 }
 
+Base.@kwdef struct QPForwCache
+    dz::Vector{Float64}
+    dλ::Vector{Float64}
+    dν::Vector{Float64}
+end
+Base.@kwdef struct QPBackCache
```
matbesancon, line 105: Do we need to define two different types for forward and backward in the QP case?

src/quadratic_diff.jl
```diff
@@ -0,0 +1,29 @@
...
+    )
+
+    LHS = create_LHS_matrix(z, λ, Q, G, h, A)
+    model.gradient_cache = QPCache(
+        problem_data,
+        λ,
+        ν,
+        LHS,
+    )
+end
```
matbesancon, line 29:
```suggestion
end

```

src/MOI_wrapper.jl
joaquimg: - [ ] It is a little drafty, I am was planning to dispatch on them.
Bu now there are separate fields so I might just merge them.

joaquimg: merged these two

joaquimg:
The key thing to do now is to map between the user facing indices for variables and constraints to the one mapping into the Matrix Forms, possible `index_map` is all I need

joaquimg:
Also, I might try using MatOI.GeometricConicForm for QP matrix building

joaquimg:
Ok, the general sketch is there, now I will work on getting this up and running.

joaquimg:
The first tests are already good.
There is still some work to do due to interface changes.

joaquimg:
This should be good to start bikeshedding on names.
Open to suggestions @matbesancon @AKS1996 @blegat 

Besides tests and docs...
I might merge some caches.
And I still have to convert TODO's into issues.

src/MOI_wrapper.jl
```diff
@@ -333,7 +338,13 @@ function get_dA(b_cache::QPForwBackCache, g_cache::QPCache, vi, ci::CI{F,S}
     dz = b_cache.dz
     λ = g_cache.inequality_duals
     dλ = b_cache.dλ
-    return λ[i] * (dλ[i] * z[j] + λ[i] * dz[j])
+    # this is the previously implemented
+    # return Diagonal(λ) * dλ * z' - λ * dz')
+    return λ[i] * (dλ[i] * z[j]) - λ[i] * dz[j]
+    # from the paper, teh correct solution should be this
+    # and thec correct fix is probably correcting the signs of the duals
```
matbesancon, line 345:
```suggestion
    # from the paper, the correct solution should be this
    # and the correct fix is probably correcting the signs of the duals
```

src/utils.jl
```diff
@@ -40,7 +40,7 @@ function create_LHS_matrix(z, λ, Q, G, h, A=nothing)
                 A         spzeros(p, m)          spzeros(p, p)]
     end
 end
-
+# TODO: this is teh transpose, check back fowr usage
```
matbesancon, line 43:
```suggestion
# TODO: this is the transpose, check back for usage
```

matbesancon:
@joaquimg there seems to be a conflict after the last merge?

joaquimg:
I have some local changes I want to push today.

joaquimg:
Current tests are passing and so are docs.

I am in favor of merging this PR ASAP so that new work can follow the MOI style to pass and access data.

Things to do before merging:
 - [ ] fix conflicts
 - [ ] add basic tests for Conic backward and QP forward

Things to be addressed in following PRs:
 - Improve coverage
 - Improve docs
 - Open issues related to the highlighted TODO's
 
 cc @matbesancon good plan?
 
 suggestions?

matbesancon:
Agreed with the plan yes!

test/qp_forward.jl
```diff
@@ -0,0 +1,41 @@
...
+
+    MOI.optimize!(model)
+
+    MOI.set(model, DiffOpt.ForwardIn{DiffOpt.ConstraintConstant}(), c2, 1.0)
+
+    DiffOpt.forward!(model)
+
+    dx = MOI.get(model, DiffOpt.ForwardOut{MOI.VariablePrimal}(), v[])
+    @test dx ≈ 1.0  atol=ATOL rtol=RTOL
+end
```
matbesancon, line 41:
```suggestion
end

```

joaquimg:
Conflicts solved. However, the recent JuMP tests use the old interface.
They were commented out.

matbesancon:
> However, the recent JuMP tests use the old interface. They were commented out

We should probably keep the JuMP tests and adapt them to the new interface, they also show how to use DiffOpt with JuMP

joaquimg:
I added a JuMP interface, first jump tests are already passing, last ones need to be updated

src/MOI_wrapper.jl
```diff
@@ -93,30 +99,445 @@ Base.@kwdef struct ConicCache
...
+```julia
+MOI.set(model, DiffOpt.ForwardIn{DiffOpt.LinearObjective}(), x)
+```
+"""
+struct ForwardIn{T} <: AbstractDiffAttribute end
+
+"""
+    ForwardOut{T}
+
+A AbstractDiffAttribute to set input data to backward differentiation, that
```
matbesancon, line 176: if it's for backward diff, why still call the attribute forward? I think I am missing the notion attached to In and Out in this context

joaquimg: It is not back, it is forward indeed. The text is wrong. Thanks for finding this!

In and Out are the input differentials and output differentials.

matbesancon:
What is still missing apart from the updated docs? (for In/Out and Forward Backward)?

AKS1996:
went through this PR. Seems I messed up big time confusing between forward and backward modes :sweat_smile: (made clear while creating `rrule`'s for DiffOpt)
Just to be clear, forward pass is problem data  `A,b,c` -> `x,y,s`
and backward pass is `dx,dy,dz` -> `dA,db,dc` right?

matbesancon:
@AKS1996 forward and backward passes are both needed for reverse diff.
In forward diff you can do both in just one forward pass `(A,b,c), (dA,db,dc) -> d_out`

> and backward pass is dx,dy,dz -> dA,db,dc right?

not exactly. The input of the backward pass is the derivative of some function wrt each output `dl/dx, dl/dy...` and the output is the derivative of the same function wrt the parameters `dl/dA, dl/db...`

joaquimg:
> What is still missing apart from the updated docs? (for In/Out and Forward Backward)?

The last few tests from JuMP are missing, we need to port these to the new interface, which is matrix agnostic as users see a model in JuMP.

Docs can be improved, but we can think about that in following PR's

> not exactly. The input of the backward pass is the derivative of some function wrt each output dl/dx, dl/dy... and the output is the derivative of the same function wrt the parameters dl/dA, dl/db...

Yeah, backward is a bit confusing, because in Boyd's papers they do say dx,dy,dz -> dA,db,dc
However, people actually want what @matbesancon said, which is mostly written in Brandon Amos' thesis.

joaquimg:
3 tests to go, should be finished tomorrow.

joaquimg:
Nightly is failing, might be upstream or dependency version related.

odow:
It's a MutableArithmetics thing. Not your fault.

joaquimg:
All tests are updated and passing as they were before this PR.
There is a bunch of TODO´s to address in the following PR´s.
I will start opening issues.

@matbesancon @be-apt would you like to take a look?

Project.toml
```diff
@@ -6,6 +6,7 @@ version = "0.1.0"
 [deps]
 BlockDiagonals = "0a1fb500-61f7-11e9-3c65-f5ef3456f9f0"
 IterativeSolvers = "42fd0dbc-a981-5370-80f2-aaf504508153"
+JuMP = "4076af6c-e467-56ae-b986-b466b2749572"
```
matbesancon, line 9: Do we need JuMP in main dependencies? Shouldn't it be a Test (and possibly docs) dep?

src/MOI_wrapper.jl
```diff
@@ -93,30 +99,445 @@ Base.@kwdef struct ConicCache
...
+    Nothing,
+    QPForwBackCache,
+    ConicForwCache,
+}
+const CACHE_BACK_TYPE = Union{
+    Nothing,
+    QPForwBackCache,
+    ConicBackCache,
+}
+# TODO bechmark access to these, ^^, type unstable type
```
matbesancon, line 135:
```suggestion
# TODO benchmark access to these, ^^, type unstable type
```

src/quadratic_diff.jl
```diff
@@ -0,0 +1,47 @@
...
+        MOI.get.(model.optimizer, MOI.ConstraintDual(), le_con_sv_idx),
+    )
+    append!(
+        λ,
+        MOI.get.(model.optimizer, MOI.ConstraintDual(), ge_con_sv_idx),
+    )
+    ν = MOI.get.(model.optimizer, MOI.ConstraintDual(), eq_con_idx)
+    append!(
+        ν,
+        MOI.get.(model.optimizer, MOI.ConstraintDual(), eq_con_sv_idx)
```
matbesancon, line 34:
```suggestion
        MOI.get.(model.optimizer, MOI.ConstraintDual(), eq_con_sv_idx),
```

src/utils.jl
```diff
@@ -40,7 +40,7 @@ function create_LHS_matrix(z, λ, Q, G, h, A=nothing)::AbstractMatrix{Float64}
                 A         spzeros(p, m)          spzeros(p, p)]
     end
 end
-
+# TODO: this is the transpose, check back for usage
```
matbesancon, line 43: what does this mean here?

src/utils.jl
```diff
@@ -246,3 +276,201 @@ function _slack_dual_vectors(model)
...
+function MOI.set(
+    m::MOI.Utilities.CachingOptimizer,
+    attr::AbstractDiffAttribute,
+    index::MOI.Index,
+    value,
+)
+    if m.state == MOI.Utilities.ATTACHED_OPTIMIZER
+        optimizer_index = m.model_to_optimizer_map[index]
+        optimizer_value = MOI.Utilities.map_indices(m.model_to_optimizer_map, value)
+        # if m.mode == AUTOMATIC
```
matbesancon, line 291: should this be removed? Also should we add a docstring to justify this method?

src/utils.jl
```diff
@@ -246,3 +276,201 @@ function _slack_dual_vectors(model)
...
+end
+function MOI.set(
+    b::MOI.Bridges.AbstractBridgeOptimizer,
+    attr::AbstractDiffAttribute,
+    index::MOI.Index,
+    index2::MOI.Index,
+    value,
+)
+    MOI.set(b.model, attr, index, index2, value)
+end
```
matbesancon, line 476:
```suggestion
end

```

test/conic_backward.jl
```diff
@@ -0,0 +1,53 @@
...
+    @test x ≈ ones(3) atol=ATOL rtol=RTOL
+
+    MOI.set(model, DiffOpt.BackwardIn{MOI.VariablePrimal}(), X[1], 1.0)
+
+    DiffOpt.backward!(model)
+
+    db = MOI.get(model, DiffOpt.BackwardOut{DiffOpt.ConstraintConstant}(), c)
+
+    @test db ≈ [-1.0]  atol=ATOL rtol=RTOL
+    # TODO: unifor sign
```
matbesancon, line 52: what does this one mean?

src/utils.jl
joaquimg: This was copied (and simplified) from MOI´s caching optimizer´s code. (https://github.com/jump-dev/MathOptInterface.jl/blob/f585f81caddd54b8cc9843c263ee89b28ebc9a9e/src/Utilities/cachingoptimizer.jl#L662)
We need this to allow `AbstractDiffAttribute` to flow through this layer.
Referencing MOI's code in a comment is a good idea.

We might want add back the `if m.mode == AUTOMATIC` block.

I think we can stick with the `ATTACHED_OPTIMIZER` error for simplicity.

src/utils.jl
joaquimg: This comment is about the above function.
It is used in both back and forward methods.
However, it seems to be correct for backward methods, but it seems to be transposed in the forward method.

test/conic_backward.jl
joaquimg: I believe this sign might be off.
Because increasing x should be in line with increasing b, (the constraint is x == b).
I am not an expert in backward mode though. Have to think a bit about it.
Also, I followed the paper signs but MOI conventions might be flipping some duals.

matbesancon (approved):

matbesancon:
alright we have a bunch of issues still left but they can be kept for some other time

matbesancon:
merging in few hours / tomorrow if no further objection, thanks @joaquimg !

matbesancon:
@joaquimg as a follow-up, can you open issues for the things that were still pending there?

joaquimg:
I will!

Issue #82, opened 2021-03-20
SingleVariable objective should be compatible with QPs


matbesancon:
```julia
const QP_OBJECTIVE_TYPES = Union{
    MathOptInterface.ScalarAffineFunction{Float64},
    MathOptInterface.ScalarQuadraticFunction{Float64},
}
```
This makes LPs with `MOI.SingleVariable` conic programs, which is a bit overkill

matbesancon:
this errors, because calls the conic interface:
```julia
    model = direct_model(diff_optimizer(Clp.Optimizer))
    MOI.set(model, MOI.Silent(), true)

    @variable(model, x[1:1])

    @objective(model, Min, x[1])
    @constraint(model, x[1] ≥ 0)
    @constraint(model, x[1] ≥ 3)

    optimize!(model)

    # obtain gradients
    grads = backward(JuMP.backend(model), ["G", "h"], [1.0])
```

This fixes the problem:
```julia
    @objective(model, Min, 1.1 * x[1])
```

joaquimg:
This should be fixed in the #81 rewrite

matbesancon:
OK great. I had a quick fix in an upcoming PR but a more systematic one will be better

joaquimg:
fixed by #83

PR #83, opened 2021-03-20
add JuMP tests

matbesancon:
this copies the examples and tests from MOI to JuMP.
Most examples worked out of the box, this adds a quick fix to #82 

The conic examples are broken, with unsupported constraints, not sure why exactly (`@test_broken` added)


joaquimg (commented):
Nice additions,
I am proposing a more MOI style interface in #81, which might make jump queries more natural and similar to querying classic results from primals and duals

joaquimg:
We can merge this asap since tests already pass.
I can try to integrate in the other
PR.

matbesancon:
Yes the interface is still clumsly, and the conic form is still not doable in the same style. Agreed to merge this

PR #88, opened 2021-04-21
editing SVM analysis example 

be-apt:
using JuMP instead of MOI (less code, more understandable)
fixed a minor typo


be-apt:
@joaquimg any clue why is this failing? the script `unit_example.jl` was working earlier i guess

matbesancon:
just re-running to see if the error on nightly is still there

matbesancon:
@be-apt when using `julia` in quote blocks in documentation, it will just create a colorized snippet. What we want is actually running the code, that's why the blocks are named

be-apt:
i guess we can drop `example 1` vs `julia` thing once https://github.com/jump-dev/DiffOpt.jl/issues/93 is closed

matbesancon:
@be-apt for now let's keep the `example 1` blocks to have the examples built, we can move to Literate later

matbesancon:
example failing here

be-apt:
closing this. just a duplicate of #95

PR #90, opened 2021-04-25
bug fixes in QP backend

be-apt:
Wanted to emulate something like this example: https://github.com/cvxgrp/cvxpylayers/blob/master/examples/torch/tutorial.ipynb
Made some small fixes
Here's the demo: https://github.com/be-apt/jump-gsoc-2020/blob/master/sensitivity-analysis-ridge.ipynb

### DIFF
--- a/src/MOI_wrapper.jl
+++ b/src/MOI_wrapper.jl
@@ -347,13 +347,13 @@ end
 
 function MOI.get(model::Optimizer,
     ::ForwardIn{QuadraticObjective}, vi1::VI, vi2::VI)
-    tuple = ifelse(vi1.value <= vi2.value, (vi1, vi2), (vi2, vi1))
-    return get(model.input_cache.dQ, tuple, 0.0)
+    idx = ifelse(vi1.value <= vi2.value, (vi1, vi2), (vi2, vi1))
+    return get(model.input_cache.dQ, idx, 0.0)
 end
 function MOI.set(model::Optimizer,
     ::ForwardIn{QuadraticObjective}, vi1::VI, vi2::VI, val)
-    tuple = ifelse(vi1.value <= vi2.value, (vi1, vi2), (vi2, vi1))
-    model.input_cache.dQ[tuple] = val
+    idx = ifelse(vi1.value <= vi2.value, (vi1, vi2), (vi2, vi1))
+    model.input_cache.dQ[idx] = val
     return
 end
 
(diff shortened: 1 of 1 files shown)


be-apt:
I see a major feature requirement (thanks to cvxpylayer), i.e. ability to track gradients (if we can somehow pass a TrackedArray instead of an array in a JuMP model). This allows fine-grained control of differentiating a function, which can be a composition of  variables and tracked problem data (such as a loss) directly.

matbesancon:
Normally we shouldn't need tracked arrays. This is the way it was done in
Tracker, the former AD system for Flux. It's more a mechanism for
propagating gradient information rather than a feature in itself.

src/MOI_wrapper.jl
```diff
@@ -347,12 +347,12 @@ end
 
 function MOI.get(model::Optimizer,
     ::ForwardIn{QuadraticObjective}, vi1::VI, vi2::VI)
-    tuple = ifelse(index(vi1) <= index(vi2), (vi1, vi2), (vi2, vi1))
+    tuple = ifelse(vi1.value <= vi2.value, (vi1, vi2), (vi2, vi1))
```
matbesancon, line 350: why changing this one?

src/MOI_wrapper.jl
```diff
@@ -871,7 +871,7 @@ function _fill_quad_Q(model, dQv, dQi, dQj, index_map)
     for ((vi1, vi2), val) in dict_dQ
         i = index_map[vi1].value
         j = index_map[vi2].value
-        for (vi, val) in dict
+        for (vi, val) in dict_dQ
```
matbesancon, line 874: ok this is weird that this wasn't caught by tests before

matbesancon: @joaquimg should there be a test where `dict_dQ` gets used?

matbesancon:
@be-apt since this PR seems to be focused on the small bug fixes, maybe we should rename it and open something for the regression examples?

src/MOI_wrapper.jl
be-apt: it was failing on my device, on Julia 1.0

be-apt: image

src/MOI_wrapper.jl
be-apt: yep. i guess with more examples common, such bugs (if any) will surface

src/MOI_wrapper.jl
matbesancon: one thing to note: `tuple` is also a Julia function in Base, maybe try to use a name that is not already defined to avoid tricky bugs

be-apt:
is this good to go?

matbesancon:
rename the `tuple` variables to something else and good to go

PR #91, opened 2021-05-01
added ChainRules examples with forward and backward

matbesancon:
Copying the unit commitment example with ChainRules.

We depend on ChainRulesCore through SetDistances, so I added it as a dependency

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -5,6 +5,7 @@ version = "0.1.0"
 
 [deps]
 BlockDiagonals = "0a1fb500-61f7-11e9-3c65-f5ef3456f9f0"
+ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
 IterativeSolvers = "42fd0dbc-a981-5370-80f2-aaf504508153"
 JuMP = "4076af6c-e467-56ae-b986-b466b2749572"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
--- a/examples/chainrules.jl
+++ b/examples/chainrules.jl
@@ -4,11 +4,23 @@ using DiffOpt
 using Test
 using ChainRulesCore
 
-# This script creates the JuMP problem for a small unit commitment instance
+# This script creates the JuMP model for a small unit commitment instance
+# represented in a solution map function taking parameters as arguments and returning
+# the optimal solution as output.
+
+# The derivatives of this solution map can then be expressed in ChainRules semantics
+# and implemented using DiffOpt
 
 ATOL=1e-4
 RTOL=1e-4
 
+"""
+Solution map of the problem using parameters:
+- `load1_demand, load2_demand` for the demand of two nodes
+- `gen_costs` is the vector of generator costs
+- `noload_costs` is the vector of fixed activation costs of the generators,
+and returning the optimal output power `p`.
+"""
 function unit_commitment(load1_demand, load2_demand, gen_costs, noload_costs; model = Model(() -> diff_optimizer(Clp.Optimizer)))
     ## Problem data
     unit_codes = [1, 2] # Generator identifiers
@@ -58,14 +70,10 @@ function unit_commitment(load1_demand, load2_demand, gen_costs, noload_costs; mo
     return JuMP.value.(p.data)
 end
 
-unit_commitment([1.0, 1.2, 1.4, 1.6], [1.0, 1.2, 1.4, 1.6], [1000.0, 1500.0], [500.0, 1000.0])
+@show unit_commitment([1.0, 1.2, 1.4, 1.6], [1.0, 1.2, 1.4, 1.6], [1000.0, 1500.0], [500.0, 1000.0])
 
-
-load1_demand = [1.0, 1.2, 1.4, 1.6]
-load2_demand = [1.0, 1.2, 1.4, 1.6]
-gen_costs = [1000.0, 1500.0]
-noload_costs = [500.0, 1000.0]
-
+# Forward differentiation rule for the solution map of the unit commitment problem
+# taking in input perturbations on the input parameters and returning perturbations propagated to the result
 function ChainRulesCore.frule((_, Δload1_demand, Δload2_demand, Δgen_costs, Δnoload_costs), ::typeof(unit_commitment), load1_demand, load2_demand, gen_costs, noload_costs)
     model = Model(() -> diff_optimizer(Clp.Optimizer))
     pv = unit_commitment(load1_demand, load2_demand, gen_costs, noload_costs, model=model)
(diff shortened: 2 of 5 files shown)


be-apt:
this is great. i've been stuck at finding the difference b/t frules and rrules in implementation. now its much clear

matbesancon:
@oxinabox if you want to take a look

Project.toml
```diff
@@ -5,6 +5,7 @@ version = "0.1.0"
 
 [deps]
 BlockDiagonals = "0a1fb500-61f7-11e9-3c65-f5ef3456f9f0"
+ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
```
oxinabox, line 8: Why not just as a docs dependency?

oxinabox:
Looks good.
It would be good to have more narrative around this to have in the docs.
Like https://jump.dev/DiffOpt.jl/dev/sensitivity-analysis-svm/ has with plots.
Possibly someone on our power systems team can help suggesting things that are interesting to show.

Project.toml
matbesancon: 1. we already depend on it transitively through MathOptSetDistances
2. At some point we will use ChainRulesCore instead of querying the projection Jacobian from MathOptSetDistances

examples/chainrules.jl
```diff
@@ -0,0 +1,138 @@
...
+function unit_commitment(load1_demand, load2_demand, gen_costs, noload_costs; model = Model(() -> diff_optimizer(Clp.Optimizer)))
+    ## Problem data
+    unit_codes = [1, 2] # Generator identifiers
+    load_names = ["Load1", "Load2"] # Load identifiers
+    n_periods = 4 # Number of time periods
+    Pmin = Dict(1 => fill(0.5, n_periods), 2 => fill(0.5, n_periods)) # Minimum power output (pu)
+    Pmax = Dict(1 => fill(3.0, n_periods), 2 => fill(3.0, n_periods)) # Maximum power output (pu)
+    RR = Dict(1 => 0.25, 2 => 0.25) # Ramp rates (pu/min)
+    P0 = Dict(1 => 0.0, 2 => 0.0) # Initial power output (pu)
+    D = Dict("Load1" => load1_demand, "Load2" => load2_demand) # Demand
```
raphaelsaavedra, line 33:
```suggestion
    D = Dict("Load1" => load1_demand, "Load2" => load2_demand) # Demand (pu)
```

Just to have units for everything for completeness

examples/chainrules.jl
```diff
@@ -0,0 +1,138 @@
...
+    Cp = Dict(1 => gen_costs[1], 2 => gen_costs[2]) # Generation cost coefficient ($/pu)
+    Cnl = Dict(1 => noload_costs[1], 2 => noload_costs[2]) # No-load cost ($)
+
+    ## Variables
+    @variable(model, 0 <= u[g in unit_codes, t in 1:n_periods] <= 1) # Commitment
```
raphaelsaavedra, line 38: Would be good to have a comment on the fact we're using the linear relaxation for the sake of clarity

examples/chainrules.jl
```diff
@@ -0,0 +1,138 @@
...
+
+@show unit_commitment([1.0, 1.2, 1.4, 1.6], [1.0, 1.2, 1.4, 1.6], [1000.0, 1500.0], [500.0, 1000.0])
+
+# Forward differentiation rule for the solution map of the unit commitment problem
+# taking in input perturbations on the input parameters and returning perturbations propagated to the result
+function ChainRulesCore.frule((_, Δload1_demand, Δload2_demand, Δgen_costs, Δnoload_costs), ::typeof(unit_commitment), load1_demand, load2_demand, gen_costs, noload_costs)
+    model = Model(() -> diff_optimizer(Clp.Optimizer))
+    pv = unit_commitment(load1_demand, load2_demand, gen_costs, noload_costs, model=model)
+    energy_balance_cons = model[:energy_balance_cons]
+    MOI.set.(model, DiffOpt.ForwardIn{DiffOpt.ConstraintConstant}(), energy_balance_cons, [d1 + d2 for (d1, d2) in zip(Δload1_demand, Δload1_demand)])
```
raphaelsaavedra, line 81: This might be a dumb question because I'm not familiar with DiffOpt and AD in general, but what we're doing here exactly and why is it specifically for the energy balance constraint?

raphaelsaavedra (commented):
Thanks for the work, looks really nice. Left some minor comments and a question.

examples/chainrules.jl
matbesancon: we are setting the perturbation for the right-hand-side of the energy balance constraints, which is the sum of perturbations of each of the quantities in the RHS (in that case sum of load demands)

examples/chainrules.jl
```diff
@@ -0,0 +1,156 @@
...
+RTOL=1e-4
+
+"""
+Solution map of the problem using parameters:
+- `load1_demand, load2_demand` for the demand of two nodes
+- `gen_costs` is the vector of generator costs
+- `noload_costs` is the vector of fixed activation costs of the generators,
+and returning the optimal output power `p`.
+"""
+function unit_commitment(load1_demand, load2_demand, gen_costs, noload_costs; model = Model(() -> diff_optimizer(Clp.Optimizer)))
```
oxinabox, line 24: I think we can just do
```suggestion
function unit_commitment(load1_demand, load2_demand, gen_costs, noload_costs; model = Clp.Optimizer)
```

examples/chainrules.jl
```diff
@@ -0,0 +1,156 @@
...
+    return JuMP.value.(p.data)
+end
+
+@show unit_commitment([1.0, 1.2, 1.4, 1.6], [1.0, 1.2, 1.4, 1.6], [1000.0, 1500.0], [500.0, 1000.0])
+
+# Forward differentiation rule for the solution map of the unit commitment problem
+# taking in input perturbations on the input parameters and returning perturbations propagated to the result
+function ChainRulesCore.frule((_, Δload1_demand, Δload2_demand, Δgen_costs, Δnoload_costs), ::typeof(unit_commitment), load1_demand, load2_demand, gen_costs, noload_costs)
+    # creating the UC model with a DiffOpt optimizer wrapper around Clp
+    model = Model(() -> diff_optimizer(Clp.Optimizer))
```
oxinabox, line 82: We need to also pass the kwargs in: so probably can do:
```suggestion
function ChainRulesCore.frule((_, Δload1_demand, Δload2_demand, Δgen_costs, Δnoload_costs), ::typeof(unit_commitment), load1_demand, load2_demand, gen_costs, noload_costs; model=Model(Clp.Optimizer))
    # creating the UC model with a DiffOpt optimizer wrapper around Clp
    model = Model(() -> diff_optimizer(optimizer(model)))
```

examples/chainrules.jl
```diff
@@ -0,0 +1,156 @@
...
+Δload1_demand = 0 * load1_demand .+ 0.1
+Δload2_demand = 0 * load2_demand .+ 0.2
+Δgen_costs = 0 * gen_costs .+ 0.1
+Δnoload_costs = 0 * noload_costs .+ 0.4
+@show (pv, Δpv) = ChainRulesCore.frule((nothing, Δload1_demand, Δload2_demand, Δgen_costs, Δnoload_costs), unit_commitment, load1_demand, load2_demand, gen_costs, noload_costs)
+
+# Reverse-mode differentiation of the solution map
+# The computed pullback takes a seed for the optimal solution `̄p` and returns
+# derivatives wrt each input parameter.
+function ChainRulesCore.rrule(::typeof(unit_commitment), load1_demand, load2_demand, gen_costs, noload_costs; model = Model(() -> diff_optimizer(Clp.Optimizer)))
```
oxinabox, line 127: we should write this assuming the Model they use isn't a `diff_optimizer` model.
And we should unwrap and rewrap the model

examples/chainrules.jl
matbesancon: one issue here is that `backend(.)` only gives you the top-most backend layer, so you add two layers of CachingOptimizer

examples/chainrules.jl
matbesancon: unwrapping and rewrapping the model results in the same issue as above

Issue #117, opened 2021-06-22
Affine objective path never taken


blegat:
This will always return a quadratic objective, even if the objective is affine (by converting it to quadratic), so the first clause if-else after that is never used.
https://github.com/jump-dev/DiffOpt.jl/blob/b6ec288ff9fabb039cc6e86f67af4a9d0c2e24da/src/quadratic_diff.jl#L272

Issue #116, opened 2021-06-22
Possibly incorrect extraction of affine terms with quadratic objective


blegat:
I don't understand how can this be correct:
https://github.com/jump-dev/DiffOpt.jl/blob/b6ec288ff9fabb039cc6e86f67af4a9d0c2e24da/src/quadratic_diff.jl#L292
We take the coefficients and we don't even look at the variable indices

Issue #115, opened 2021-06-22
Speed up construction of sparse matrix representation of objective


blegat:
Currently, we use `setindex!` on sparse arrays which is slow:
https://github.com/jump-dev/DiffOpt.jl/blob/b6ec288ff9fabb039cc6e86f67af4a9d0c2e24da/src/quadratic_diff.jl#L273-L293

PR #118, opened 2021-06-22
Fix and speed up quadratic objective conversion to sparse array format

blegat:
Closes https://github.com/jump-dev/DiffOpt.jl/issues/117
Closes https://github.com/jump-dev/DiffOpt.jl/issues/116
Closes https://github.com/jump-dev/DiffOpt.jl/issues/115


blegat:
Merged as it was needed for https://github.com/jump-dev/DiffOpt.jl/pull/119

Issue #114, opened 2021-06-22
Attributes as Variable, Constraint and Model attributes


blegat:
The attributes we define are not subtypes of variable, constraint or model attributes because the value passed is described sometimes with a number, sometimes with a variable and a number and sometimes with 2 variables and a numbers.
The variables should be mapped with the `index_map` when going through the `CachingOptimizer` layer for instance.
The fact that we pass several things can be handled by wrapping them in a `struct` and the fact that the variables need to be mapped can be by defining a `map_indices`.
In fact, we could even reuse the `MOI.ScalarAffineTerm` and `MOI.ScalarQuadraticTerm`.
Moreover, adding derivatives for for multiple coefficients could be done with `MOI.ScalarAffineFunction` and `MOI.ScalarQuadraticFunction`.
In fact, this would even work with variable bridges out of the box with `bridged_function`.
If you pass a `MOI.ScalarAffineFunction` containing, say,  `2x + 3y` and `x` is bridged into `a - b` then `bridged_function` will transform this into `2a - 2b + 3y` which is actually what we want.

matbesancon:
Was this one closed by the last PR?

blegat:
Yes, closed by https://github.com/jump-dev/DiffOpt.jl/pull/128

PR #121, opened 2021-06-23
BackwardOut{...Objective} -> BackwardOutObjective

### LINKED ISSUE #122: Typo in conic getter for BackwardOut objective
`dQ_end_i` should be one of the two operands of the last line of
https://github.com/jump-dev/DiffOpt.jl/blob/2a373d07b1c45b02b5d03c9699ee6b23f70297bc/src/diff_opt.jl#L307-L309

### LINKED ISSUE #123: Bug due to mismatch with dual sign convention
In this line:
https://github.com/jump-dev/DiffOpt.jl/blob/6a7450152c8f7fcfc3db792eb969b19c70e70ac1/src/diff_opt.jl#L702
This should be `- dA' * ν` even if it is `+ dA' * ν` because `ν`  is the MOI dual which is the opposite as the dual used in the paper.
We should probably reverse `ν` when we get the dual so that we don't have these complications.
This sign was correctly reversed here:
https://github.com/jump-dev/DiffOpt.jl/blob/6a7450152c8f7fcfc3db792eb969b19c70e70ac1/src/diff_opt.jl#L467-L472

### LINKED ISSUE #125: Missing transpose
In https://github.com/jump-dev/DiffOpt.jl/blob/6a7450152c8f7fcfc3db792eb969b19c70e70ac1/src/diff_opt.jl#L708-L710, we should be using `LHS'` because LHS is defined as the matrix defined in (7) in the paper which is the transpose of the matrix defined in (6) that we need.

blegat:
Follow up of https://github.com/jump-dev/DiffOpt.jl/pull/119
See https://github.com/jump-dev/DiffOpt.jl/issues/114

Closes #122
Closes #123
Closes #125


src/diff_opt.jl
```diff
@@ -296,10 +310,37 @@ function MOI.set(model::Optimizer, ::BackwardInVariablePrimal, vi::VI, val)
     return
 end
 
-function MOI.get(model::Optimizer, ::BackwardOut{LinearObjective}, vi::VI)
-    return _get_dc(model, vi)
+function MOI.get(model::Optimizer, ::BackwardOutObjective)
```
joaquimg, line 313: You rarely want all the sensitivities.
If you have a model with many variables this will waste a lot of time while the user might just care about a single variable coefficient.

joaquimg (commented):
The new attributes are fine. But we should be able to query results sparsely.

matbesancon:
that's a good point yup, that was the motivation for the MOI attributes in the first place

src/diff_opt.jl
```diff
@@ -180,6 +177,7 @@ MOI.is_set_by_optimize(::ForwardOutVariablePrimal) = true
 
 A `MOI.AbstractVariableAttribute` to set input data to backward
 differentiation, that is, problem solution.
+T
```
matbesancon, line 180: typo?

src/jump_moi_overloads.jl
```diff
@@ -1,6 +1,127 @@
...
+    return MOI.set(model, attr, JuMP.moi_function(func))
+end
+
+abstract type AbstractLazyScalarFunction <: MOI.AbstractScalarFunction end
+standard_form(func::Union{MOI.SingleVariable,MOI.ScalarAffineFunction,MOI.ScalarQuadraticFunction}) = func
+function Base.isapprox(func1::AbstractLazyScalarFunction, func2::MOI.AbstractScalarFunction; kws...)
+    return isapprox(standard_form(func1), standard_form(func2); kws...)
+end
+
+struct RankOneQuadraticFunction{T, VT<:AbstractVector{T}} <: AbstractLazyScalarFunction
```
matbesancon, line 12: docstring

matbesancon:
this introduces several elements that are not yet explained? What's the MOItoJuMP and why needed?

blegat:
> The new attributes are fine. But we should be able to query results sparsely.

This is adressed in the last commit.

> this introduces several elements that are not yet explained? What's the MOItoJuMP and why needed?

Yes, I should add docstrings. It seems more consistent to return a JuMP.AbstractJuMPScalar than a MOI.AbstractScalarFunction if the getter is called on a JuMP model. The fact that it holds a JuMP model allows it to check that the variables indices given belong to the right model, handle conversion to JuMP functions and printing.

src/diff_opt.jl
```diff
@@ -189,42 +187,39 @@ MOI.set(model, DiffOpt.BackwardInVariablePrimal(), x)
...
 
+"""
+    BackwardOutObjective
+
+A `MOI.AbstractModelAttribute` to get output data to backward differentiation,
+that is, problem input data.
+
+```julia
+MOI.get(model, DiffOpt.BackwardOutObjective)
+```
```
joaquimg, line 198: it would be great to have two line showing:
- get sensitivity of linear part of obj
- get sensitivity of quad part of obj

src/jump_moi_overloads.jl
```diff
@@ -1,6 +1,127 @@
...
+function standard_form(func::MOItoJuMP)
+    return JuMP.jump_function(func.model, standard_form(func.func))
+end
+function JuMP.function_string(mode, func::MOItoJuMP)
+    return JuMP.function_string(mode, standard_form(func))
+end
+
+function MOI.get(model::JuMP.Model, attr::BackwardOutObjective)
+    func = MOI.get(JuMP.backend(model), attr)
+    return JuMP.jump_function(model, func)
```
joaquimg, line 124: do we still need the function below this one?

blegat:
In view of the bug https://github.com/jump-dev/DiffOpt.jl/issues/122 being unnoticed by the tests and that I wanted to be sure that we get it right, I thought we could have having integration tests checking for the scalar products between forward and backward diffs, which are unlikely to be correct if there are buts.
I added such tests in the latest commit. It didn't at first and after some debugging (which was so intense I missed the beginning of JuMP-dev talk :/), I found also
https://github.com/jump-dev/DiffOpt.jl/issues/125
https://github.com/jump-dev/DiffOpt.jl/issues/123
I took me some time to nail https://github.com/jump-dev/DiffOpt.jl/issues/123 down ^^

src/jump_moi_overloads.jl
blegat: Yes, we still have BackwardIn and BackwardOut for the constraints.

src/diff_opt.jl
blegat: Done

src/jump_moi_overloads.jl
blegat: Done

blegat:
Any objection to merge ?

src/jump_moi_overloads.jl
```diff
@@ -1,6 +1,210 @@
...
+        ],
+        func.constant,
+    )
+end
+function standard_form(func::VectorScalarAffineFunction{T}) where {T}
+    return convert(MOI.ScalarAffineFunction{T}, func)
+end
+
+"""
+    struct MatrixScalarQuadraticFunction{T, VT, MT} <: MOI.AbstractScalarFunction
```
matbesancon, line 79: do we really need the full struct in the docstring?

joaquimg:
Go for it!
Next step constraint attributes?

joaquimg (approved):

joaquimg (approved):

src/jump_moi_overloads.jl
blegat: Fixed in latest commit

matbesancon (approved):

PR #134, opened 2021-07-16
Allow setting explicit problem class

blegat:
One design decision here would be: do we want to have this as type parameter of `Optimizer` or as a field ? The advantage of a field is that it can be set with optimizer attributes and change at will while even if it may not make much sense to change it once a few constraints have already been passed and bridges have been chosen in view of the program class. But a user may want to change the class, empty and copy again, that would make sense.
Let me know what you think

### DIFF
--- a/src/diff_opt.jl
+++ b/src/diff_opt.jl
@@ -249,7 +249,7 @@ different program class:
 2) Conic Program (CP): linear objective and conic constraints.
 
 `AUTOMATIC` which means that the class will be automatically selected given the
-problem data: if any constraint is conic, QP is used and CP is used otherwise.
+problem data: if any constraint is conic, CP is used and QP is used otherwise.
 See [`ProgramClass`](@ref).
 """
 @enum ProgramClassCode QUADRATIC CONIC AUTOMATIC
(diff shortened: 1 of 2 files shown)


src/diff_opt.jl
```diff
@@ -240,9 +240,25 @@ MOI.get(model, DiffOpt.BackwardOutConstraint(), ci)
...
+"""
+    @enum ProgramClassCode QUADRATIC CONIC AUTOMATIC
+
+Program class used by DiffOpt. DiffOpt implements differentiation of two
+different program class:
+1) Quadratic Program (QP): quadratic objective and linear constraints and
+2) Conic Program (CP): linear objective and conic constraints.
+
+`AUTOMATIC` which means that the class will be automatically selected given the
+problem data: if any constraint is conic, QP is used and CP is used otherwise.
```
joaquimg, line 252: `if any constraint is conic, CP is used and QP is used otherwise.`

joaquimg (approved):

matbesancon:
I agree with the current design, having it as a field and attribute is more flexible

PR #136, opened 2021-07-20
SVM Analysis using Literate

be-apt:


docs/src/examples/sensitivity-analysis-svm.jl
```diff
@@ -0,0 +1,248 @@
+# # Sensitivity Analysis of SVM
+
+#md # [![](https://img.shields.io/badge/GitHub-100000?style=for-the-badge&logo=github&logoColor=white)](@__REPO_ROOT_URL__/examples/sensitivity-analysis-svm.jl)
+#md # [![](https://img.shields.io/badge/show-nbviewer-579ACA.svg)](@__NBVIEWER_ROOT_URL__/generated/sensitivity-analysis-svm.ipynb)
+
+
+# This notebook illustrates sensitivity analysis of data points in an [Support Vector Machine](https://en.wikipedia.org/wiki/Support-vector_machine) (inspired from [@matbesancon](http://github.com/matbesancon)'s [SimpleSVMs](http://github.com/matbesancon/SimpleSVMs.jl).)
```
blegat, line 7: an -> a

docs/src/examples/sensitivity-analysis-svm.jl
```diff
@@ -0,0 +1,248 @@
...
+
+# For reference, Section 10.1 of https://online.stat.psu.edu/stat508/book/export/html/792 gives an intuitive explanation of what does it means to have a sensitive hyperplane or data point. The general form of the SVM training problem is given below (without regularization):
```
blegat, line 9: "what does it means" - "what it means"

blegat (commented):
What's the reason we use MOI and not JuMP in the example ?

joaquimg:
Maybe we can even have both...

be-apt:
> What's the reason we use MOI and not JuMP in the example ?

will update the example using JuMP. that way it'll be easier to understand too

blegat:
> Maybe we can even have both...

Yes, and the difference between the two can be quite small, see https://github.com/jump-dev/MathOptInterface.jl/issues/1472#issuecomment-881150059

docs/src/examples/sensitivity-analysis-svm.jl
```diff
@@ -0,0 +1,226 @@
...
+# Add the constraints.
+
+@constraint(
+    model, 
+    1.0*l ∈ MOI.Nonnegatives(N),
+)
+@constraint(
+    model, 
+    cons, 
+    y.*(X*w .+ b) + l.-1 ∈ MOI.Nonnegatives(N),
```
blegat, line 61: Use spaces around binary operators

.gitignore
```diff
@@ -3,6 +3,7 @@
 *.jl.mem
 docs/build/
 docs/site/
+docs/src/examples/*.md
```
joaquimg, line 6: why ignore these?

docs/src/examples/sensitivity-analysis-svm.jl
```diff
@@ -0,0 +1,226 @@
...
+# ## Define and solve the SVM
+
+# Import the libraries.
+
+
+using SCS, DiffOpt, LinearAlgebra, JuMP
+import Random, Plots
+
+
+# Construct separatable, non-trivial data points.
```
joaquimg, line 33: separable?

docs/src/examples/sensitivity-analysis-svm.jl
```diff
@@ -0,0 +1,226 @@
...
+
+N = 100
+D = 2
+Random.seed!(62)
+X = vcat(randn(N ÷ 2, D), randn(N ÷ 2, D) .+ [4.0, 1.5]')
+y = append!(ones(N ÷ 2), -ones(N ÷ 2));
+
+
+# Let's define the model
```
joaquimg, line 42: Let's initialize a special model that  can understand sensitivities

docs/src/examples/sensitivity-analysis-svm.jl
```diff
@@ -0,0 +1,226 @@
...
+# Add the constraints.
+
+@constraint(
+    model, 
+    1.0 * l ∈ MOI.Nonnegatives(N),
+)
+@constraint(
+    model, 
+    cons, 
+    y .* (X * w .+ b) + l .- 1 ∈ MOI.Nonnegatives(N),
```
joaquimg, line 61: y .* (X * w .+ b) + l .- 1 >= 0 ?

docs/src/examples/sensitivity-analysis-svm.jl
```diff
@@ -0,0 +1,226 @@
...
+
+@variable(model, l[1:N])
+@variable(model, w[1:D])
+@variable(model, b);
+
+# Add the constraints.
+
+@constraint(
+    model, 
+    1.0 * l ∈ MOI.Nonnegatives(N),
```
joaquimg, line 56: `l .>= 0` ?

joaquimg:
Looking good!

.gitignore
be-apt: these are generated by literate right? the source literate file and markdown files in the same folder might be redundant
if you want, we can keep them in the docs folder - that way apart from https://jump.dev/DiffOpt.jl/dev, one can access docs here too

docs/src/examples/sensitivity-analysis-svm.jl
be-apt: changed

docs/src/examples/sensitivity-analysis-svm.jl
be-apt: i get
```
MethodError: no method matching set(::Model, ::DiffOpt.ForwardInConstraint, ::Vector{ConstraintRef{Model, MathOptInterface.ConstraintIndex{MathOptInterface.ScalarAffineFunction{Float64}, MathOptInterface.GreaterThan{Float64}}, ScalarShape}}, ::MathOptInterface.VectorAffineFunction{Float64})
```

be-apt:
any changes on this?

docs/src/examples/sensitivity-analysis-svm.jl
blegat: You should use broadcast: `MOI.set.`

blegat:
See https://github.com/jheinen/GR.jl/issues/422 for how to fix the docs build

be-apt:
@matbesancon any idea why this might be failing
```
Differentiating simple PSD back: Error During Test at /home/runner/work/DiffOpt.jl/DiffOpt.jl/test/conic_backward.jl:10
  Got exception outside of a @test
  MethodError: *(::Adjoint{Float64, Vector{Float64}}, ::BlockDiagonals.BlockDiagonal{Float64, Matrix{Float64}}) is ambiguous. Candidates:
    *(M::AbstractMatrix, B::BlockDiagonals.BlockDiagonal{T, V} where V<:AbstractMatrix{T}) where T in BlockDiagonals at /home/runner/.julia/packages/BlockDiagonals/Nvfvo/src/base_maths.jl:108
    *(x::Adjoint{T, <:AbstractVector} where T, A::AbstractMatrix) in LinearAlgebra at /opt/hostedtoolcache/julia/nightly/x64/share/julia/stdlib/v1.8/LinearAlgebra/src/matmul.jl:113
  Possible fix, define
    *(::Adjoint{T, <:AbstractVector} where T, ::BlockDiagonals.BlockDiagonal{T, V} where V<:AbstractMatrix{T}) where T
```

thought https://github.com/invenia/BlockDiagonals.jl/pull/55 fixed it

matbesancon:
I suspect there was a dependency blocking us, the BlockDiagonals version used was 0.1.6 which might be from before the fix

matbesancon:
@be-apt this is fixed, was probably one dependency blocking. Could you check why the docs are failing and whether the change in the other PR would fix it? If so we should probably have a PR fixing the docs action first, and then see to merge this

matbesancon:
I'm still not sure what is wrong with the documentation step here, one plot seems not to be produced, causing an error

matbesancon:
@odow were there some gotchas when you switch the documentation to Literate for JuMP?
This seems to build the markdown fine but errors when making the docs themselves

odow:
Not sure. This looks like it's a problem with the plotting, not Documenter or Literate.

blegat:
I had the same error on https://github.com/JuliaPolyhedra/Polyhedra.jl/pull/265 and fixed it in https://github.com/JuliaPolyhedra/Polyhedra.jl/pull/265/commits/5db5c2d80c92632be371b172fa99542c95969f2d.
See https://github.com/jheinen/GR.jl/issues/422#issuecomment-888953237

matbesancon:
perfect thanks @blegat

PR #138, opened 2021-08-01
Autotuning Ridge Example

be-apt:
using literate.jl
a lot of files are common to #136, only the example source code differing


joaquimg:
why docs/src/examples/sensitivity-analysis-svm.jl is being removed?

matbesancon:
good question, just re-added

docs/src/examples/autotuning-ridge.jl
```diff
@@ -0,0 +1,194 @@
...
+# ```math
+# \begin{equation}
+# \min_{w} \quad \frac{1}{2n} \sum_{i=1}^{n} (y_{i} - w^T x_{i})^2 + \alpha \| w \|_2^2
+# \end{equation}
+# ```
+
+# where 
+# - `x`, `y` are the data points
+# - `w` constitutes weights of the regressing line
+# - `α` is the only hyperparameter - regularization constant
```
matbesancon, line 18:
```suggestion
# - `α` is the only hyperparameter acting on regularization
```
I wouldn't say constant if we are indirectly optimizing over it

docs/src/examples/autotuning-ridge.jl
```diff
@@ -0,0 +1,194 @@
...
+    l = N ÷ 2  # test train split
+    return X[1:l, :], X[l+1:N, :], Y[1:l], Y[l+1:N]
+end
+
+X_train, X_test, Y_train, Y_test = create_problem(800, 30, 4);
+
+
+# Define a helper function for regression
+
+function fitRidge(X,Y,α)
```
matbesancon, line 57:
```suggestion
function fit_ridge(X, Y, α)
```

docs/src/examples/autotuning-ridge.jl
```diff
@@ -0,0 +1,194 @@
...
+end
+
+# Solve the problem for several values of α
+
+αs = [0.0, 1e-3, 2e-3, 5e-3, 1e-2, 2e-2, 5e-2, 7e-2, 2e-1, 3e-1, .5, .7, 1.0]
+Rs = Float64[]
+mse = Float64[]
+
+for α in αs
+    _, _, _, w_train = fitRidge(X_train, Y_train, α)
```
matbesancon, line 84:
```suggestion
    _, _, _, w_train = fit_ridge(X_train, Y_train, α)
```

docs/src/examples/autotuning-ridge.jl
```diff
@@ -0,0 +1,194 @@
...
+end
+
+
+# Plot the gradient ∂l/∂α
+
+∂l_∂αs = Float64[]
+N, D = size(X_train)
+
+for α in αs
+    model, w, _, ŵ = fitRidge(X_train, Y_train, α)
```
matbesancon, line 138:
```suggestion
    model, w, _, ŵ = fit_ridge(X_train, Y_train, α)
```

docs/src/examples/autotuning-ridge.jl
```diff
@@ -0,0 +1,194 @@
...
+    prev_mse = 1e7
+    curr_mse = 1e6
+    
+    α_s = Float64[]
+    mse = Float64[]
+    
+    iter=0
+    while curr_mse-10 < prev_mse && iter < max_iters
+        iter += 1
+        model, w, _, ŵ = fitRidge(X_train, Y_train, α)
```
matbesancon, line 170:
```suggestion
        model, w, _, ŵ = fit_ridge(X_train, Y_train, α)
```

docs/src/examples/autotuning-ridge.jl
```diff
@@ -0,0 +1,194 @@
...
+"""
+function descent(α, max_iters=25)
+    prev_mse = 1e7
+    curr_mse = 1e6
+    
+    α_s = Float64[]
+    mse = Float64[]
+    
+    iter=0
+    while curr_mse-10 < prev_mse && iter < max_iters
```
matbesancon, line 168:
```suggestion
    while curr_mse - 10 < prev_mse && iter < max_iters
```

docs/src/examples/autotuning-ridge.jl
```diff
@@ -0,0 +1,194 @@
...
+    
+    return α_s, mse
+end
+
+ᾱ, msē = descent(1.0);
+
+# Visualize gradient descent and convergence 
+
+plot(log.(αs), mse, label="MSE", xaxis = ("α"))
+plot!(log.(ᾱ), msē, label="G.D. for α", lw = 2)
```
matbesancon, line 194:
```suggestion
plot!(log.(ᾱ), msē, label="G.D. for α", lw = 2)

```

docs/src/examples/autotuning-ridge.jl
```diff
@@ -0,0 +1,194 @@
...
+    return 1-(u/v)
+end
+
+# Create a non-trivial, noisy regression dataset
+
+function create_problem(N, D, noise)
+    w = rand(D) 
+    X = rand(N, D) 
+    
+    Y = X*w .+ noise*randn(N) # if noise=0, then there is no need of regularization and alpha=0 will give the best R2 score
```
blegat, line 46: If `y` is a vector, use a lower case

docs/src/examples/autotuning-ridge.jl
```diff
@@ -0,0 +1,194 @@
...
+
+    N, D = size(X)
+    
+    @variable(model, w[1:D]>=-10)
+    set_optimizer_attribute(model, MOI.Silent(), true)
+    
+    @objective(
+        model,
+        Min,
+        sum((Y - X*w).*(Y - X*w))/(2.0*N) + α*(w'w),
```
blegat, line 68: You take a scalar product in two different ways here. Doing `dot(Y - X*w, Y - X*w)` and `dot(w, w)` might be clearer

docs/src/examples/autotuning-ridge.jl
```diff
@@ -0,0 +1,194 @@
...
+    )
+
+    optimize!(model)
+
```
blegat, line 72: Check the status and show a warning or error if it's not what's expected

docs/src/examples/autotuning-ridge.jl
```diff
@@ -0,0 +1,194 @@
...
+end
+
+plot(αs, ∂l_∂αs, label="∂l/∂α",  xaxis = ("α"))
+
+
+
+# Define helper function for Gradient Descent
+
+"""
+    start from initial value of regularization constant
```
blegat, line 156: Don't indent the docstring. Add `descent(α, max_iters=25)` indented above it.

docs/src/examples/autotuning-ridge.jl
```diff
@@ -0,0 +1,194 @@
...
+using Statistics
+using OSQP
+using JuMP
+using Plots
+import Random
+using LinearAlgebra
+
+
+"""
+Return the coefficient of determination R2 of the prediction.
```
blegat, line 30: Add `    R2(y_true, y_pred)` before this line

matbesancon:
There is something wrong with the model, the loss derivative wrt alpha is incorrect, as it is always positive, leading alpha always closer to 0

matbesancon:
@be-apt not sure what is wrong, but the gradient computation does not reflect the MSE function

docs/src/examples/autotuning-ridge.jl
```diff
@@ -0,0 +1,198 @@
...
+
+plot(log.(αs), mse, label="MSE", xaxis = ("log(α)"))
+
+# Define the derivative of the model with respect to the parameter α
+function ∇model(model, X_train, w, ŵ, α)
+    N, D = size(X_train)
+    dw = zeros(D)
+    ∂w_∂α = zeros(D)
+
+    for i in 1:D
```
blegat, line 114: Could you clarify what is done here? Since the parameter is alpha, I would expect the input to be `w*w`

docs/src/examples/autotuning-ridge.jl
```diff
@@ -0,0 +1,198 @@
...
+        dw[i] = 1 #set
+
+        MOI.set(
+            model, 
+            DiffOpt.ForwardInObjective(), 
+            MOI.ScalarQuadraticFunction(
```
blegat, line 120: Build the function at the JuMP level using `w`, `+`, `*`,...

docs/src/examples/autotuning-ridge.jl
matbesancon: yes, I changed it to that but hasn't improved things yet

matbesancon: the parameter is indeed just `w * w`, there was a weird mechanism which just multiplied by 1.0

matbesancon:
the example is still broken for now but fixed various things. One major point: the taken gradient was that over the training set, in which case the optimal alpha will always be 0

PR #150, opened 2021-11-11
Upgrading DiffOpt to MOI 0.10

matbesancon:
Still a lot of errors I haven't figured out yet

- [x] Replace MatrixOptInterface by MOI.Utilities.MatrixOfConstraints

Closes https://github.com/jump-dev/DiffOpt.jl/pull/150


matbesancon:
starts to tackle #150

blegat:
There is an issue on Julia v1.0:

* LazyArrays is v0.16 or earlier for Julia v1.0, these version don't support StaticArrays v1.0
* MathOptSetDistances (MOSD) requires StaticArrays v1.0

We have 2 options: either relax the requirement in MOSD or drop support for old Julia versions.

matbesancon:
we should just ditch Julia 1.0 for DiffOpt, it's not directly a solver and not registered yet, 1.6 can become the new lower bound

matbesancon:
I set julia 1.6 as the lower bound for now on this

blegat (approved):
Looks good to merge, let's track the issue of the dirty flag in https://github.com/jump-dev/DiffOpt.jl/issues/152 and not keep this PR open for this.

joaquimg (approved):

Issue #152, opened 2021-11-11
Allow setting and getting without optimize in between


blegat:
See https://github.com/jump-dev/JuMP.jl/issues/2797

PR #154, opened 2021-11-13
Fix getting ForwardOutVariablePrimal with JuMP v0.22

blegat:
Closes https://github.com/jump-dev/DiffOpt.jl/issues/152

### DIFF
--- a/src/jump_moi_overloads.jl
+++ b/src/jump_moi_overloads.jl
@@ -14,9 +14,29 @@ function MOI.get(model::JuMP.Model, attr::BackwardOutObjective)
 end
 
 function MOI.get(model::JuMP.Model, attr::BackwardOutConstraint, con_ref::JuMP.ConstraintRef)
-    check_belongs_to_model(con_ref, model)
+    JuMP.check_belongs_to_model(con_ref, model)
     moi_func = MOI.get(JuMP.backend(model), attr, JuMP.index(con_ref))
     return JuMP.jump_function(model, moi_func)
+end
+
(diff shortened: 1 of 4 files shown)


src/jump_moi_overloads.jl
```diff
@@ -14,11 +14,31 @@ function MOI.get(model::JuMP.Model, attr::BackwardOutObjective)
...
 
 function MOI.get(model::JuMP.Model, attr::BackwardOutConstraint, con_ref::JuMP.ConstraintRef)
-    check_belongs_to_model(con_ref, model)
+    JuMP.check_belongs_to_model(con_ref, model)
     moi_func = MOI.get(JuMP.backend(model), attr, JuMP.index(con_ref))
     return JuMP.jump_function(model, moi_func)
 end
 
+# FIXME Workaround for https://github.com/jump-dev/JuMP.jl/issues/2797
+function _moi_get_result(model::MOI.ModelLike, args...)
```
odow, line 23: Why'd you add `_moi_get_result`?

blegat: I wanted to call it from line 39 but I forgot. It seems best to do it to keep a consistent behavior. A solution is still needed so the status shouldn't be `OPTIMIZE_NOT_CALLED`.

odow:
So this wasn't quite the fix I suggested. It now allows. you to get `ForwardOutVariablePrimal` if the model was modified. I suggested instead that setting `ForwardInConstraint` would skip the `is_model_dirty` flag.

PR #95, opened 2021-05-23
More examples

be-apt:

PR #164, opened 2021-12-04
Review examples

joaquimg:
replaces #95


joaquimg:
@matbesancon should we keep:
```
"../examples/unit_example.jl"
"../examples/chainrules.jl"
```
?

I removed the custom SVM example with flux because it did not use DiffOpt, the derivatives were set to zero.

I might bring back "solve conic" without cvxpy and sensible results

matbesancon:
> "../examples/unit_example.jl"
"../examples/chainrules.jl"

you mean given that we have a corresponding tutorial? Or do you think they are too redundant with one another?

joaquimg:
> you mean given that we have a corresponding tutorial? Or do you think they are too redundant with one another?

Yes, I think they are a bit redundant.
I would vote for keeping just `chainrule_unit` the tutorial that is a nice and complete example.

joaquimg:
Good to merge on my side.
If you like to chat about it let me know @matbesancon , we might be better talking offline if you think there are too many changes.

matbesancon (approved):

docs/src/examples/sensitivity-analysis-ridge.jl
```diff
@@ -0,0 +1,132 @@
...
+Random.seed!(42)
+
+N = 100
+
+w = 2 * abs(randn())
+b = rand()
+X = randn(N)
+Y = w * X .+ b + 0.8 * randn(N);
+
+# The helper method `fitRidge` defines and solves the corresponding model.
```
matbesancon, line 56: `fit_ridge` would be more idiomatic.
Also some more comments maybe between the code parts?

docs/src/examples/sensitivity-analysis-ridge.jl
```diff
@@ -130,3 +136,5 @@ p = Plots.scatter(
 mi, ma = minimum(X), maximum(X)
 Plots.plot!(p, [mi, ma], [mi * ŵ + b̂, ma * ŵ + b̂], color = :red, label = "")
 
+# Note the points in the extremes of the line segment are larger because
+# moving those points can affect more the angular coefficient of the line.
```
matbesancon, line 140:
```suggestion
# moving those points has more effect on the angular coefficient of the line.

```

Issue #174, opened 2021-12-07
Error in README example


joaquimg:
```julia

using JuMP
import DiffOpt
import GLPK

model = JuMP.Model(() -> DiffOpt.diff_optimizer(GLPK.Optimizer))

@variable(model, x)
@constraint(
  model,
  cons,
  x >= 3,
)
@objective(
  model,
  Min,
  2x,
)

optimize!(model) # solve

MOI.set.(  # set pertubations / gradient inputs
    model, 
    DiffOpt.BackwardInVariablePrimal(),
    x,
    1.0,
)

DiffOpt.backward(model) # differentiate

grad_exp = MOI.get(   # -3x+1
    model,
    DiffOpt.BackwardOutConstraint(),
    cons
)
```

According to the readme comment, the last line should return `-3x+1`.

However, it returns `-3x-1`.

joaquimg:
If we have: `(1+y) * x = 3`
That is equivalent to `x = 3/(1+y)`
The derivative is: `dx/dy = -3/(1+y)^2` at `y = 0` we have `dx/dy = -3`

If we have: `x = 3 + y`
The derivative is: `dx/dy =+ 1` 

Hence, the comment if correct, but result seems incorrect.

This is probably just related to standard forms.

joaquimg:
On the other hand we might be understating the sensitivity as:

Normalize the constraint: `x  - 3 + y = 0`
Then isolate `x`
The derivative is: `dx/dy = -1`

In this case, the result is correct. (and comment in wrong).

joaquimg:
It makes sense that we are doing the latter because we are passing perturbations as functions, so we care about normalized constraints which extend cleanly to Conic programming.

PR #177, opened 2021-12-09
Add explanation and fix typo in some tutorials

JinraeKim:
I've followed some tutorials, and especially in comparison between manual QP differentiation and using DiffOpt,
I thought it would be better to add some explanation why `0.0 * index(x[1])` is used and what `-1.0` stands for `cause it's confusing considering the value of `h = -1.0`.

### DIFF
--- a/docs/src/examples/custom-relu.jl
+++ b/docs/src/examples/custom-relu.jl
@@ -135,4 +135,4 @@ accuracy(test_X, test_Y)
 
 # Note that the accuracy is low due to simplified training.
 # It is possible to increase the number of samples `N`,
-# the number of epochs `epoch` and teh conectivity `inner`.
+# the number of epochs `epoch` and the connectivity `inner`.
--- a/docs/src/examples/custom-relu.jl
+++ b/docs/src/examples/custom-relu.jl
@@ -64,7 +64,7 @@ function ChainRulesCore.rrule(
             obj_exp = MOI.get(
                 model,
                 DiffOpt.BackwardOutObjective()
-            ) # return grdiente wrt objective function parameters
+            ) # return gradient wrt objective function parameters
             dy[:, i] = JuMP.coefficient.(obj_exp, x) # coeff of `x` in -2x'y
             dy[:, i] = -2 * dy[:, i]
         end
--- a/docs/src/examples/matrix-inversion-manual.jl
+++ b/docs/src/examples/matrix-inversion-manual.jl
@@ -143,8 +143,10 @@ MOI.set(
     model,
     DiffOpt.ForwardInConstraint(),
     cons[1],
-    0.0 * index(x[1]) - 1.0,
+    0.0 * index(x[1]) - 1.0,  # the tangent of the ConstraintFunction, i.e., ∂(Gx - h)/∂h = -1
 )
+
+# Note that `0.0 * index(x[1])` is used to make its type `MOI.AbstractScalarFunction`.
 
 # Compute derivatives
 
@@ -156,7 +158,7 @@ dx = MOI.get.(
     model,
     DiffOpt.ForwardOutVariablePrimal(),
     x,
-)
+)  # ∂x/∂h
 
 using Test                                  #src
 @test dx ≈ [0.25 ,0.75] atol=1e-4 rtol=1e-4 #src
--- a/docs/src/examples/matrix-inversion-manual.jl
+++ b/docs/src/examples/matrix-inversion-manual.jl
@@ -147,6 +147,7 @@ MOI.set(
 )
 
 # Note that `0.0 * index(x[1])` is used to make its type `typeof(0.0 * index(x[1]) - 1.0) <: MOI.AbstractScalarFunction`.
+# To indicate different direction to get directional derivative, users should replace `0.0 * index(x[1]) - 1.0` as the form of `dG*x - dh`, where `dG` and `dh` correspond to the elements of direction vectors along `G` and `h` axes, respectively.
 
 # Compute derivatives
 
--- a/docs/src/examples/matrix-inversion-manual.jl
+++ b/docs/src/examples/matrix-inversion-manual.jl
@@ -143,7 +143,7 @@ MOI.set(
     model,
     DiffOpt.ForwardInConstraint(),
     cons[1],
-    0.0 * index(x[1]) - 1.0,  # the tangent of the ConstraintFunction, i.e., ∂(Gx - h)/∂h = -1
+    0.0 * index(x[1]) - 1.0,  # to indicate the direction vector to get directional derivatives
 )
 
 # Note that `0.0 * index(x[1])` is used to make its type `typeof(0.0 * index(x[1]) - 1.0) <: MOI.AbstractScalarFunction`.
--- a/docs/src/examples/matrix-inversion-manual.jl
+++ b/docs/src/examples/matrix-inversion-manual.jl
@@ -44,8 +44,8 @@
 # ```math
 # \begin{gather}
 #  \begin{bmatrix} 
-#      Q & g^T \\
-#      \lambda^* g & g x^* - h
+#      Q & G^T \\
+#      \lambda^* G & G x^* - h
 #  \end{bmatrix}
 #  \begin{bmatrix} 
 #      dx \\
--- a/docs/src/examples/matrix-inversion-manual.jl
+++ b/docs/src/examples/matrix-inversion-manual.jl
@@ -146,7 +146,7 @@ MOI.set(
     0.0 * index(x[1]) - 1.0,  # the tangent of the ConstraintFunction, i.e., ∂(Gx - h)/∂h = -1
 )
 
-# Note that `0.0 * index(x[1])` is used to make its type `MOI.AbstractScalarFunction`.
+# Note that `0.0 * index(x[1])` is used to make its type `typeof(0.0 * index(x[1]) - 1.0) <: MOI.AbstractScalarFunction`.
 
 # Compute derivatives
 


JinraeKim:
I'm not familiar with generating documentation so I modified some files, which look like some source files to generate the corresponding documentation.

odow:
> I modified some files

You chose wisely.

You can build the docs by running:
```Julia
cd ~/.julia/dev/DiffOpt
julia --project=docs docs/make.jl
```
If it complains about not being able to find DiffOpt, then
```Julia
cd ~/.julia/dev/DiffOpt
julia --project=docs

julia> ] dev .
julia> exit()

julia --project=docs docs/make.jl
```

JinraeKim:
Should I build new documentation and add it to this PR?

As I'm not familiar with generating docs, I'm worried about it :)

odow:
> Should I build new documentation and add it to this PR?

Nope. There is an automatic CI job that builds the documentation and publishes a new version of the website:
https://github.com/jump-dev/DiffOpt.jl/runs/4465766227?check_suite_focus=true

JinraeKim:
@odow If you don't mind, can you answer one the following question?
Then, where can I find and check the new version of automatically generated website?

odow:
Ah. Good question. If you had commit rights to DiffOpt, then you could view it at 
https://github.com/jump-dev/DiffOpt.jl/previews/PR177. 

But since you're an external contributor you don't have permission to publish a new version:
image

Instead, we just make sure that the CI job succeeds.

When this PR is merged it will rebuild the docs.

You don't need to do anything else for this PR. But I'll let someone who works on DiffOpt.jl review it in more detail and then merge.

JinraeKim:
@odow 
I just wondered if I can check the auto-generated docs.
Thanks!

odow:
Yeah for external contributors you have to build them locally to check. 

But also, it's not the end of the world if there is a small mistake that slips through the PR. You can always wait for the docs to build on `master` and then fix any mistakes in another PR. In general, I try to encourage (without much success) people to make documentation pull requests. If you just touch files in `/docs` you can't break the code so nothing bad can happen if you make a mistake.

docs/src/examples/matrix-inversion-manual.jl
```diff
@@ -143,9 +143,11 @@ MOI.set(
     model,
     DiffOpt.ForwardInConstraint(),
     cons[1],
-    0.0 * index(x[1]) - 1.0,
+    0.0 * index(x[1]) - 1.0,  # the tangent of the ConstraintFunction, i.e., ∂(Gx - h)/∂h = -1
```
joaquimg, line 146: Not exactly.
This is actually a perturbation on the problem. That is the direction that the derivative will be computed. In this case the direction is only following the "h[1]" axis and nothing else. There is one axis for each parameter of the problem "h[1]", "G[1,1]", "G[1,2]" and so on.

Then, the get ForwardOutVariablePrimal will return the derivative of 'x' in that direction

joaquimg (changes requested):
Thanks a lot!

There is on imprecision that we should fix before merging.

docs/src/examples/matrix-inversion-manual.jl
JinraeKim: Thank you for your comment.
I'm really sorry that I don't get it clearly.

Can you explain why the argument should be `0.0 * index(x[1]) - 1.0` and how it changes if different parameter is taken for calculation of derivatives?

JinraeKim: Apart from that, I think that the output of `get ForwardOutVariablePrimal` returns `dx/dh` in the tutorial, right?

joaquimg: Derivatives in Rn require a direction, see:
https://en.wikipedia.org/wiki/Directional_derivative

that argument is defining that direction to be (1 * h[1], 0*G[1,1], 0*G[1,2], ...).

In other (very different) words, that carries the information of which vector is going to be multiplied by the Jacobian: (1 * dh[1], 0*dG[1,1], 0*dG[1,2], ...)

joaquimg: see also: https://math.stackexchange.com/questions/3191003/directional-derivative-and-jacobian-matrix

docs/src/examples/custom-relu.jl
```diff
@@ -135,4 +135,4 @@ accuracy(test_X, test_Y)
 
 # Note that the accuracy is low due to simplified training.
 # It is possible to increase the number of samples `N`,
-# the number of epochs `epoch` and teh conectivity `inner`.
\ No newline at end of file
+# the number of epochs `epoch` and teh conectivity `inner`.
```
matbesancon, line 138:
```suggestion
# the number of epochs `epoch` and the connectivity `inner`.
```

docs/src/examples/matrix-inversion-manual.jl
JinraeKim: Thank you, @joaquimg.
But my question is, in that line, "how to let the code know the direction" for directional derivative calculation.
That is,
I don't understand how the line `0.0 * index(x[1]) - 1.0` indicates that "the direction is (1, 0, 0, ...) with axes of (h[1], G[1, 1], G[1, 2], ...)".

JinraeKim: What if I want to decide the direction as (0, 1, 0) with axes of (h[1], G[1, 1], G[1, 2])?

joaquimg: In that case you would do:
1.0x[1] + 0.0

cons[1] means that you are selecting the first constraint and x[1] means the coefficient of the first variable: that is G[1,1].

The 1.0 multiplying x[1] means 1 in that direcation.

JinraeKim: Well, I see.
Then, you mean that when considering the constraint `Gx - h <=0`, the direction of `G[1, i]` corresponds to `x[i]` and `h` corresponds to the constant...?
Also,
1) the `-1.0` in `0.0 * index(x[1]) - 1.0` corresponds to "1 in h-direction", or "-h"-direction? (it should be "h-direction" in context, probably due to the negative sign convention of `h` in `Gx - h <= 0`.
2) `0.0 * index(x[1])` is still to make the type a AbstractScalarSomething? (I forgot the name, sorry)

JinraeKim: EDIT:
I'm not sure but from my observation that the `0.0 * index(x[1]) - 1.0` corresponds to `dG * x - dh`; probably that's why @joaquimg referred to as "directional derivative".
This can be found in some source codes like [how to get `dh`](https://github.com/jump-dev/DiffOpt.jl/blob/cf808b04c122a73500cdd4db9bc269b63ee3b709/src/diff_opt.jl#L689).

Please inform me if I misunderstood.

joaquimg: the sign on the constant is a bit non-trivial, see my own reflections here: https://github.com/jump-dev/DiffOpt.jl/issues/174
might help...

JinraeKim: @joaquimg #174 provides a very good insight of how it works.
It would be better to read it before reading the source codes... 👍

JinraeKim: @joaquimg 
Can you check whether I understood it correctly?

The constraint function is basically the form of `f(x, G, h_) = Gx + h_` internally,
so to get the directional derivative `h`-axis (with unit vector), we have to flip the sign of the constant as we consider the constraint of `f(x, G, -h) = Gx - h`.
Also, to get directional derivatives of other axes, in the case of `G[1, i]`, just passing `x[i]` into where `0.0 * index(x[1]) - 1.0` is placed provides the corresponding directional derivative (with unit vector).
Multiplication by scalar replaces the unit vector condition intuitively.

joaquimg: looks good!

JinraeKim, line 146:
```suggestion
    0.0 * index(x[1]) - 1.0,  # to indicate the direction vector to get directional derivatives
```

docs/src/examples/matrix-inversion-manual.jl
```diff
@@ -143,9 +143,11 @@ MOI.set(
...
 )
 
+# Note that `0.0 * index(x[1])` is used to make its type `typeof(0.0 * index(x[1]) - 1.0) <: MOI.AbstractScalarFunction`.
```
JinraeKim, line 149:
```suggestion
# Note that `0.0 * index(x[1])` is used to make its type `typeof(0.0 * index(x[1]) - 1.0) <: MOI.AbstractScalarFunction`.
# To indicate different direction to get directional derivative, users should replace `0.0 * index(x[1]) - 1.0` as the form of `A*x - b`, where `A` and `b` correspond to the elements of direction vectors along `G` and `h` axes, respectively.
```

JinraeKim (commented):
@joaquimg
I modified the comments related to directional derivatives.

docs/src/examples/matrix-inversion-manual.jl
joaquimg: looks good for me

docs/src/examples/matrix-inversion-manual.jl
joaquimg: I would avoid using A and b, since they are used to describe equality constraints.
maybe `dG * x - dh`?

JinraeKim: IMO, it may also be confusing as in #174 we will finally get `dx/dy` where `y` denotes perturbation itself.
That is, if we use the notation as `dG` and `dh`, we will get something like `dx/ddh`.

However, I agree that still `dG` and `dh` are intuitive and convenient.

joaquimg: Forward mode is usually simpler to understand. dG and dh also follow (closely) the notation of the paper OptNet

JinraeKim: Your comment is reflected.

docs/src/examples/matrix-inversion-manual.jl
JinraeKim: For the change request, see the following commits.

joaquimg (approved):

matbesancon:
Thanks for the PR!

Issue #167, opened 2021-12-06
Refactoring with bridges


blegat:
I started a new design with https://github.com/jump-dev/DiffOpt.jl/pull/166

Currently, the user needs to either model in a purely conic way or purely QP way to have it work with DiffOpt and we don't want to restrict it to the format understood by DiffOpt as it may be different that the format needed by the optimizer and we don't want to do transformations back and forth.
The idea is that the user model should be directly given to the optimizer with minimal transformation while DiffOpt needs to standardize things into the format it understands with a separate bridging layer that it not applied to the model given to the optimizer.

So in `DiffOpt.Optimizer`, there will be
1. `optimizer`: The optimize with exactly the model the user gave (possibly some bridges there but it's transparent to DiffOpt).
2. `diff`: Either a `LazyBridgeOptimizer{QPDiff}` or a `LazyBridgeOptimizer{ConicDiff}`.

To differentiate,
1. the model is first copied from `optimizer` to `diff`, with bridges applied so that the form is exactly the conic or QP form we can work with.
2. the primal and dual values are passed from the optimizer to the diff, the bridges taking care of transforming them.
3. when the user queries or set a forward or backward info, it go through the bridges so in case that variable/constraint/objective was bridged, it will transform it just like it's a primal or dual information. This will require some changes to MOI.

- [x] Create `diff` subfield https://github.com/jump-dev/DiffOpt.jl/pull/166
- [x] Let `diff` have its own indices https://github.com/jump-dev/DiffOpt.jl/pull/179
- [x] Add a bridge layer for the conic `diff` https://github.com/jump-dev/DiffOpt.jl/pull/184
- [ ] Use matrix form for `QPDiff` https://github.com/jump-dev/DiffOpt.jl/issues/153

joaquimg:
At first glance, this looks **very** nice.

Changes in MOI make me a bit worried about to not be compatible with JuMP 1.x. I don't know what you have in mind,

On a parallel thread, but looking forward...

I imagine that many usages of this project would perform many in-place problem modifications of the problems.
We can imagine something like a gradient descent usage:
0 - Start with some value of some data like a coefficient in the objective
1 - solve model
2 - compute the derivative wrt to the coefficient
3 - update the coefficient
4 - go back to 1

I know that for many conic solvers optimizing such pipeline is not useful since model is always reloaded. But for quadratic like cplex, gurobi, express, osqp this would be very valuable.

[...]

PR #178, opened 2021-12-09
Make the ReLU example clear

### LINKED ISSUE #176: Understanding `ChainRulesCore.rrule` in custom ReLU example
Hi, developers! Again, it's me, Jinrae.

After understanding the ReLU example (#168), I'm trying to "use" DiffOpt.jl.
Unfortunately, I didn't get it properly.

If I understand the mechanism, it might be possible for me to add my understandings to the documentation (as a newbie).

# What I want to do using DiffOpt.jl
I want to make my custom neural network differentiable via Flux.jl (namely `my_NN`), which contains a convex optimisation problem.
For example, the loss function would be the mse error of
```julia
y - my_NN_optimal(x)
```
where `y` is the label, `x` is condition, and `my_NN_optimal(x) = my_NN(x, u_)` where `u_` is the minimised of `u -> my_NN(x, u)`.
To update `my_NN`'s parameters, I have to get some `rrule`.

# My understanding up to now
## What users can do using DiffOpt.jl
First, I've found in [Introduction](https://jump.dev/DiffOpt.jl/dev/intro/) that DiffOpt.jl can be used for differentiating the optimal solution to given parametrised convex optimisation problem with respect to problem parameters.
For example, let `x_` and `p` be the optimal solution and problem parameters, respectively.
It is possible to make an `rrule` for `dx_ / dp` using DiffOpt.jl.

## Strategy
As presented in some pages in docs, for example, [custom ReLU example](https://jump.dev/DiffOpt.jl/dev/examples/custom-relu/), I may have to add some `rrule`s (reverse-mode differentiation), which will be imported from [ChainRulesCore.jl](https://github.com/JuliaDiff/ChainRulesCore.jl).

## Where I felt stuck in
I'm struggling with the following things.
### What do `dx` and `dy` mean in [the custom ReLU example](https://jump.dev/DiffOpt.jl/dev/examples/custom-relu/)?
In the tutorial, especially where `ChainRulesCore.rrule` is defined, I cannot exactly follow what each line means.
It is probably due to my poor understanding of DiffOpt.jl and related packages, e.g., JuMP.

For example, this line means "performing optimisation" (with manually written function `matrix_relu`).
```julia
    pv = matrix_relu(y, model = model)
```
Then, what do the following lines mean? I mean,
- What is `dx`?
- I still don't understand even what `BackwardInVariablePrimal` is for.
```julia
            MOI.set.(
                model,
                DiffOpt.BackwardInVariablePrimal(),
                x,
                dx[:, i]
            ) # set sensitivities
```

Also, the following outputs `obj_exp` (seems an expression) but I don't get any clue about what it means.
```julia
            obj_exp = MOI.get(
                model,
                DiffOpt.BackwardOutObjective()
            ) # return grdiente wrt objective function parameters
```
For example, some outputs from REPL say
```julia
obj_exp = MOI.get(model, DiffOpt.BackwardOutObjective()) = -0.2500000001711336 x[1]² - 0.5 x[1]
-0.2500000001711336 x[1]² - 0.5 x[1]
```
so it is a quadratic expression of `x`. 
- What is `BackwardOutObjective` for?
- Why is it "the gradient wrt objective function parameters"?
- What's objective function parameters in this tutorial?

And some other questions:
- Which scheme do I have to choose among [two schemes](https://jump.dev/DiffOpt.jl/dev/manual/)?
- How can I readily transform my convex optimisation problem into a conic form (QP form would probably be restrictive in my case)?

Sorry for the verbose questions because of my lack of background for differentiable convex programming.
Thanks always.

JinraeKim:
Resolve #176.
Extended from #177.

I tried to make the ReLU example clear.

### DIFF
--- a/docs/src/examples/custom-relu.jl
+++ b/docs/src/examples/custom-relu.jl
@@ -51,8 +51,8 @@ function ChainRulesCore.rrule(
 ) where T
     pv = matrix_relu(y, model = model)
     function pullback_matrix_relu(dl_dx)
-        # # some value from the backpropagation (e.g., loss) is denoted by `l`
-        # # so `dl_dy` is the derivative of `l` wrt `y`
+        ## some value from the backpropagation (e.g., loss) is denoted by `l`
+        ## so `dl_dy` is the derivative of `l` wrt `y`
         x = model[:x] # load decision variable `x` into scope
         dl_dy = zeros(T, size(dl_dx))
         dl_dq = zeros(T, size(dl_dx)) # for step-by-step explanation
--- a/docs/src/examples/custom-relu.jl
+++ b/docs/src/examples/custom-relu.jl
@@ -67,8 +67,8 @@ function ChainRulesCore.rrule(
                 model,
                 DiffOpt.BackwardOutObjective()
             ) # return gradient wrt objective function parameters
-            dq[:, i] = JuMP.coefficient.(obj_exp, x) # coeff of `x` in q'x = -2y'x
-            dy[:, i] = -2 * dq[:, i]  # ∵ dq/dy = -2
+            dl_dq[:, i] = JuMP.coefficient.(obj_exp, x) # coeff of `x` in q'x = -2y'x
+            dl_dy[:, i] = -2 * dl_dq[:, i]  # ∵ dq/dy = -2
         end
         return (ChainRulesCore.NoTangent(), dy,)
     end
--- a/docs/src/examples/custom-relu.jl
+++ b/docs/src/examples/custom-relu.jl
@@ -52,8 +52,8 @@ function ChainRulesCore.rrule(
     pv = matrix_relu(y, model = model)
     function pullback_matrix_relu(dx)
         x = model[:x]
-        dy = zeros(T, size(dx))
-        dq = zeros(T, size(dx))  # for step-by-step explanation
+        dl_dy = zeros(T, size(dl_dx))
+        dl_dq = zeros(T, size(dl_dx))  # for step-by-step explanation
         for i in 1:size(y)[2]
             MOI.set.(
                 model,
--- a/docs/src/examples/custom-relu.jl
+++ b/docs/src/examples/custom-relu.jl
@@ -50,7 +50,8 @@ function ChainRulesCore.rrule(
     model = Model(() -> DiffOpt.diff_optimizer(OSQP.Optimizer))
 ) where T
     pv = matrix_relu(y, model = model)
-    function pullback_matrix_relu(dx)
+    function pullback_matrix_relu(dl_dx)
+    # some value from the backpropagation (e.g., loss) is denoted by `l`
         x = model[:x]
         dl_dy = zeros(T, size(dl_dx))
         dl_dq = zeros(T, size(dl_dx))  # for step-by-step explanation
--- a/docs/src/examples/custom-relu.jl
+++ b/docs/src/examples/custom-relu.jl
@@ -65,7 +65,7 @@ function ChainRulesCore.rrule(
                 model,
                 DiffOpt.BackwardOutObjective()
             ) # return gradient wrt objective function parameters
-            dq[:, i] = JuMP.coefficient.(obj_exp, x) # coeff of `x` in q'x = -2x'y
+            dq[:, i] = JuMP.coefficient.(obj_exp, x) # coeff of `x` in q'x = -2y'x
             dy[:, i] = -2 * dq[:, i]  # ∵ dq/dy = -2
         end
         return (ChainRulesCore.NoTangent(), dy,)
--- a/docs/src/examples/custom-relu.jl
+++ b/docs/src/examples/custom-relu.jl
@@ -138,4 +138,4 @@ accuracy(test_X, test_Y)
 
 # Note that the accuracy is low due to simplified training.
 # It is possible to increase the number of samples `N`,
-# the number of epochs `epoch` and teh conectivity `inner`.
+# the number of epochs `epoch` and the connectivity `inner`.
--- a/docs/src/examples/custom-relu.jl
+++ b/docs/src/examples/custom-relu.jl
@@ -34,7 +34,7 @@ function matrix_relu(
         @objective(
             model,
             Min,
-            x'x -2x'y[:, i]  # x' Q x + q'x with Q = I, q = -2y
+            x'x -2y[:, i]'x  # x' Q x + q'x with Q = I, q = -2y
         )
         optimize!(model)
         _x[:, i] = value.(x)
--- a/docs/src/examples/custom-relu.jl
+++ b/docs/src/examples/custom-relu.jl
@@ -51,8 +51,8 @@ function ChainRulesCore.rrule(
 ) where T
     pv = matrix_relu(y, model = model)
     function pullback_matrix_relu(dl_dx)
-        # some value from the backpropagation (e.g., loss) is denoted by `l`
-        # so `dl_dy` is the derivative of `l` wrt `y`
+        # # some value from the backpropagation (e.g., loss) is denoted by `l`
+        # # so `dl_dy` is the derivative of `l` wrt `y`
         x = model[:x] # load decision variable `x` into scope
         dl_dy = zeros(T, size(dl_dx))
         dl_dq = zeros(T, size(dl_dx)) # for step-by-step explanation
--- a/docs/src/examples/custom-relu.jl
+++ b/docs/src/examples/custom-relu.jl
@@ -30,7 +30,7 @@ function matrix_relu(
     empty!(model)
     set_silent(model)
     @variable(model, x[1:N] >= 0)
-    for i in 1:size(y)[2]
+    for i in 1:size(y, 2)
         @objective(
             model,
             Min,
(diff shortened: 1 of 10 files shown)


joaquimg:
Lets only edit the ReLU one here? we can leave the other tutorial for the other PR

JinraeKim:
> Lets only edit the ReLU one here? we can leave the other tutorial for the other PR

Ok, I'll do it soon.

JinraeKim:
Sorry but force-push seems not to work.
I have to leave now so I'll fix it later.

joaquimg:
just copy paste the original file

JinraeKim:
@joaquimg Here you are.

joaquimg (approved):

docs/src/examples/custom-relu.jl
```diff
@@ -64,9 +64,9 @@ function ChainRulesCore.rrule(
             obj_exp = MOI.get(
                 model,
                 DiffOpt.BackwardOutObjective()
-            ) # return grdiente wrt objective function parameters
-            dy[:, i] = JuMP.coefficient.(obj_exp, x) # coeff of `x` in -2x'y
-            dy[:, i] = -2 * dy[:, i]
+            ) # return gradient wrt objective function parameters
+            dq[:, i] = JuMP.coefficient.(obj_exp, x) # coeff of `x` in q'x = -2x'y
+            dy[:, i] = -2 * dq[:, i]  # ∵ dq/dy = -2
```
blegat, line 69: If `dq/dy = -2` then it should be `dy[:, i] = -dq[:, i] / 2`.

joaquimg: good catch

docs/src/examples/custom-relu.jl
```diff
@@ -34,7 +34,7 @@ function matrix_relu(
         @objective(
             model,
             Min,
-            x'x -2x'y[:, i]
+            x'x -2x'y[:, i]  # x' Q x + q'x with Q = I, q = -2y
```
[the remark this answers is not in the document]
joaquimg: this might be clearer

docs/src/examples/custom-relu.jl
joaquimg:
```suggestion
            dy[:, i] = -2dq[:, i]/2  # q = -2y, so dq/dy = -2 and dy = -dq/2
```

joaquimg (changes requested):
Following @blegat catch

docs/src/examples/custom-relu.jl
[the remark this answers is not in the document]
JinraeKim: I agreee

docs/src/examples/custom-relu.jl
JinraeKim: Nooooooo, that's why I was really confused with this package.
As I pointed out in [a comment, #178](https://github.com/jump-dev/DiffOpt.jl/issues/176#issuecomment-989469397), the `dy` actually refers to `dl/dy` for some output `l` (e.g., loss).

That is, the input of the pullback, `dx`, means `dl/dx`.
So `dy` and `dq` correspond to `dl/dy` and `dl/dq`, respectively.
Therefore, to convert `dq` to `dy`, we need this calculation:
`dl/dy = dl/dq * dq/dy`,
which is notationally opposite to the mathematical expression of `dy = -2 * dq` (sooo confusing).

JinraeKim: For example, in thie ReLU example, we will get `dl/dy`, denoted by `dy` in the code.
When letting `dx` (actually means `dl/dx`) be 1, we will get
`dl/dy = dl/dx * dx/dy = dx/dy`.

Since the activation ReLU provides a subgradient of
\in {0}, if `y < 0`
\in [0, 1], if `y = 0`
\in {1}, if `y > 0`
so whether the output is correct can be checked by applying any `y` into as
```julia
    m = 1
    d = 1
    y = ones(m, d)
    @show matrix_relu(y)
    pv, pullback_matrix_relu = _rrule(y)  # corresponds to ChainRulesCore.rrule
    dx = ones(m, d)  # given dl/dx
    pullback_matrix_relu(dx)  # dl/dy = dl/dx * dx/dy = dx/dy
```
gives us
```julia
dy[:, i] = -2 * dq[:, i] = [1.0]
```
which is correct.

JinraeKim: To make it clear, I added it to docs.

docs/src/examples/custom-relu.jl
```diff
@@ -53,6 +53,7 @@ function ChainRulesCore.rrule(
     function pullback_matrix_relu(dx)
         x = model[:x]
         dy = zeros(T, size(dx))
+        dq = zeros(T, size(dx))  # for step-by-step explanation
```
JinraeKim, line 56: Added for pre-assignment.

JinraeKim (commented):
Please added contents for detailed explanation

docs/src/examples/custom-relu.jl
joaquimg: might well be, i'd have to think more carefully.
if dy is actually `dl/dy,` we have to rename it to `dl_dy`.
same for `dq`.

joaquimg: same for `dx`

JinraeKim: It would be much better to use some different symbols (for example, \dot for forward mode and \overline for reverse mode, e.g., \dot{x}) as [ChainRulesCore.jl](https://juliadiff.org/ChainRulesCore.jl/dev/#Reverse-mode-AD-rules-(rrules)) to distinguish them from `dfoo`.

To avoid some Unicode issues, how 'bout `_x` for forward mode and `x_` for reverse mode?
That is, `_x = dx/da` for some `a`, and `x_ = dl/dx` for some `l`.

joaquimg: I dont see why `x_` is better than `dl_dx` to represent `dl/dx`. The first requires that the user reads a manual saying `x_ = dl/dx`, while the second is much close to what is meant.

JinraeKim: You're right, it's clearer. Honestly, both of them are fine to me :)
But in that case, users should know what `l` refers to as well.
So we would have to put some additional comments either which one is taken.

joaquimg: I agree with the need for additional comments in both options.

docs/src/examples/custom-relu.jl
JinraeKim, line 56:
```suggestion
        dl_dy = zeros(T, size(dl_dx))
        dl_dq = zeros(T, size(dl_dx))  # for step-by-step explanation
```

docs/src/examples/custom-relu.jl
```diff
@@ -53,6 +53,7 @@ function ChainRulesCore.rrule(
     function pullback_matrix_relu(dx)
```
JinraeKim, line 53:
```suggestion
    function pullback_matrix_relu(dl_dx)
    # some value from the backpropagation (e.g., loss) is denoted by `l`
```

docs/src/examples/custom-relu.jl
```diff
@@ -64,15 +65,24 @@ function ChainRulesCore.rrule(
             obj_exp = MOI.get(
                 model,
                 DiffOpt.BackwardOutObjective()
-            ) # return grdiente wrt objective function parameters
-            dy[:, i] = JuMP.coefficient.(obj_exp, x) # coeff of `x` in -2x'y
-            dy[:, i] = -2 * dy[:, i]
+            ) # return gradient wrt objective function parameters
+            dq[:, i] = JuMP.coefficient.(obj_exp, x) # coeff of `x` in q'x = -2y'x
+            dy[:, i] = -2 * dq[:, i]  # ∵ dq/dy = -2
```
JinraeKim, line 70:
```suggestion
            dl_dq[:, i] = JuMP.coefficient.(obj_exp, x) # coeff of `x` in q'x = -2y'x
            dl_dy[:, i] = -2 * dl_dq[:, i]  # ∵ dq/dy = -2
```

docs/src/examples/custom-relu.jl
```diff
@@ -64,15 +65,24 @@ function ChainRulesCore.rrule(
...
 
+# The line, `dy[:, i] = -2 * dq[:, i]` might be really confusing as
+# ```math
+# q = -2y.
+# ```
+# However, the `dfoo` in pullback refers to `dl/dfoo` for some output `l`, for example, a loss.
+# That is, `dy` actually means `dl/dy`, and therefore `dl/dy = dl/dq * dq/dy = dl/dq * (-2)` is correct
+# For more details, visit [Introduction, ChainRulesCore.jl](https://juliadiff.org/ChainRulesCore.jl/dev/).
+
+
```
JinraeKim, line 85:
```suggestion
# For more details about backpropagation, visit [Introduction, ChainRulesCore.jl](https://juliadiff.org/ChainRulesCore.jl/dev/).


```

docs/src/examples/custom-relu.jl
```diff
@@ -64,15 +65,24 @@ function ChainRulesCore.rrule(
...
                 model,
                 DiffOpt.BackwardOutObjective()
-            ) # return grdiente wrt objective function parameters
-            dy[:, i] = JuMP.coefficient.(obj_exp, x) # coeff of `x` in -2x'y
-            dy[:, i] = -2 * dy[:, i]
+            ) # return gradient wrt objective function parameters
+            dq[:, i] = JuMP.coefficient.(obj_exp, x) # coeff of `x` in q'x = -2y'x
+            dy[:, i] = -2 * dq[:, i]  # ∵ dq/dy = -2
         end
         return (ChainRulesCore.NoTangent(), dy,)
```
JinraeKim, line 72:
```suggestion
        return (ChainRulesCore.NoTangent(), dl_dy,)
```

JinraeKim (commented):
@joaquimg
`dfoo` is replaced by `dl_dfoo` in pullback.

docs/src/examples/matrix-inversion-manual.jl
```diff
@@ -159,7 +159,7 @@ dx = MOI.get.(
     model,
     DiffOpt.ForwardOutVariablePrimal(),
     x,
-)  # ∂x/∂h
+)
 
 using Test                                  #src
 @test dx ≈ [0.25 ,0.75] atol=1e-4 rtol=1e-4 #src
```
JinraeKim, line 165:
```suggestion
@test dl_dx ≈ [0.25 ,0.75] atol=1e-4 rtol=1e-4 #src
```
Fix [the test failure](https://github.com/jump-dev/DiffOpt.jl/runs/4508768666?check_suite_focus=true)

JinraeKim:
@joaquimg 
I'm confused with why rebase-and-forced-push branch shows so many file changes including #167 ...
Did I messed up?

JinraeKim:
It seems good :)

Issue #170, opened 2021-12-07
Move access to MNIST from Flux to MLDatasets


joaquimg:
Currently we get the warning:
```
┌ Warning: Flux's datasets are deprecated, please use the package MLDatasets.jl
└ @ Flux.Data C:\Users\joaquimgarcia\.julia\packages\Flux\BPPNj\src\data\Data.jl:16
```

PR #184, opened 2021-12-26
Differentiate conic model through bridges

blegat:
See https://github.com/jump-dev/DiffOpt.jl/issues/167


JinraeKim:
> See #167
> 
> ~Tests are failing because of #170~

I wonder if this bridge automatically transforms convex optimisation problems into corresponding conic problems to perform the differentiation through the optimisation problem. Is it right?
+) is this bridge automatically gonna warn if a given problem is non-convex? Or, isn't it gonna work?

blegat:
DCP will be a bridge once https://github.com/jump-dev/MathOptInterface.jl/issues/846 is done

JinraeKim:
> DCP will be a bridge once [jump-dev/MathOptInterface.jl#846](https://github.com/jump-dev/MathOptInterface.jl/issues/846) is done

I'll take a look at it. Thank you!

PR #188, opened 2022-01-05
Fix dG for QP

blegat:
Looking at (8) of https://arxiv.org/pdf/1703.00443.pdf, the previous implementation didn't seem correct, let's see what the tests say.


blegat:
There was an error in the tests as well in `test/utils.jl`, which is how this got unnoticed.
In fact, the error was detected by the tests in `test/jump.jl` and they were marked as broken for that reason. Now are passing so we now get the same results than QPTH.
@joaquimg @matbesancon @akshay326 Can one of you review this ? We should probably make a new release as well as it fixes correctness of the results.

matbesancon:
I was not familiar with the lazy combination but I trust you on the correctness.
Many of the test results were recorded and hard-copied. It makes sense to modify these

joaquimg:
Nice catch.
A broken test fixed. and the other is now correctly matching QPTH.

I agree that we should merge and release asap.

PR #210, opened 2022-04-19
Renaming

### LINKED ISSUE #208: Rename Backward to Reverse
What's the correct terminology ? The classical terminology is forward mode AD vs reverse mode AD, as in, `frule` and `rrule` so we should have `Reverse...` attributes and `DiffOpt.reverse` instead of `Backward...` and `DiffOpt.backward`.

matbesancon:
Closes #208


matbesancon:
ping @blegat @joaquimg
We can bikeshed the names, it's just a global search-and-replace now
I added the former attributes in `src/deprecated.jl`

blegat:
See my comment in https://github.com/jump-dev/DiffOpt.jl/issues/208#issuecomment-1102589302

matbesancon:
done in last commit

matbesancon:
All done @blegat

PR #216, opened 2022-04-25
Renaming

### LINKED ISSUE #215: Renaming QPDiff and ConicDiff
It's inconsistent to have `QP` and `Conic`. We should probably have `QP` and `CP`. Moreover, `DiffOpt.QPDiff` and `DiffOpt.CPDiff` are redundant, what about renaming them `DiffOpt.QP` and `DiffOpt.CP` ?

matbesancon:
Closes #215


src/quadratic_diff.jl
```diff
@@ -281,7 +281,7 @@ function reverse_differentiate!(model::QPDiff)
     dλ = partial_grads[nv+1:nv+nineq]
     dν = partial_grads[nv+nineq+1:end]
 
-    model.back_grad_cache = QPForwBackCache(dz, dλ, dν)
+    model.back_grad_cache = QPForwRevCache(dz, dλ, dν)
```
joaquimg, line 284: QP here is a bit inconsistent with naming elsewhere.

matbesancon:
increased the tolerance in the tests to fit that of HiGHS

src/quadratic_diff.jl
matbesancon: changed

PR #219, opened 2022-04-26
Refactoring into submodules

blegat:
The MOI wrapper wasn't yet aligned with the fact that we now have bridges in top of the DiffOpt models.

Closed https://github.com/jump-dev/DiffOpt.jl/issues/220


matbesancon:
Why introduce submodules? Overall I find it to quickly complicate both internal and external development, I know this is common in MOI but DiffOpt is small enough to contain everything

blegat:
We have this structure where for QP and CP, we have a `Form`, `Model`, `Cache`, ...
We need to give them different names to avoid clashes so we are adding references to `Quadratic` and `Conic` to differentiate them but it's a bit messy. The modules provide a namespace which makes it a lot cleaner.
Second, the DiffOpt optimizer is modular in the sense that it could work with any `DiffOpt.AbstractModel` subtypes that show how to differentiate for a particular model. We advertise it in the paper and the fact we can refactor cleanly with modules is a consequence of it.

Could you elaborate on why it makes things complicated ? I agree that it makes it complicated if these things are not independent but in this case, they are quite independent. We want to have a clean interface that someone can implement to add new `DiffOpt.AbstractModel`. Having these module makes this interface apparent and forces us to keep a clean one.

matbesancon:
Major hassle with submodules is the `import ..Foo or import .Bar` but I see the point that it's not too relevant here

blegat:
Yes, since the inclusion in the modules shorten the names, the name of the modules + name of the object does not increase too much in length

src/DiffOpt.jl
```diff
@@ -19,47 +18,22 @@ const MOSD = MathOptSetDistances
...
-include("conic_diff.jl")
-include("quadratic_diff.jl")
 include("moi_wrapper.jl")
 include("jump_moi_overloads.jl")
 
 include("bridges.jl")
 
+include("QuadraticProgram/QuadraticProgram.jl")
+include("ConicProgram/ConicProgram.jl")
+function add_all_model_constructors(model)
```
matbesancon, line 31: docstring for this function?

src/moi_wrapper.jl
```diff
@@ -165,41 +132,18 @@ function MOI.supports(model::Optimizer, attr::MOI.AbstractModelAttribute)
...
-function MOI.supports_constraint(
-    model::Optimizer,
-    ::Type{F},
-    ::Type{S}
-) where {
-    F <: SUPPORTED_VECTOR_FUNCTIONS,
-    S <: SUPPORTED_VECTOR_SETS,
-}
+    model::Optimizer, ::Type{F}, ::Type{S},
+) where {F<:MOI.AbstractFunction,S<:MOI.AbstractSet}
```
joaquimg, line 141: we might need to check both optimizer and the underlying diff_models

joaquimg (approved):

matbesancon:
@blegat can you check the two comments above, then good to go

matbesancon:
bump

src/moi_wrapper.jl
matbesancon: ping @blegat ?

src/DiffOpt.jl
blegat: Done

src/moi_wrapper.jl
blegat: I'd rather add special cases for the attributes that are used with the `DiffOpt.AbstractModel`, that's what I just did in the last commit

matbesancon:
@blegat good to go I think? We should tag a breaking release right after

PR #229, opened 2023-02-11
Abstract linear solving method instead of \

matbesancon:


src/QuadraticProgram/QuadraticProgram.jl
```diff
@@ -9,6 +9,8 @@ import LazyArrays
 import IterativeSolvers
 
 import DiffOpt
+import LinearSolve
+const LS = LinearSolve
```
blegat, line 13: Let's not introduce `LS` if we use it only 3 times, it just decreases readability IMO

src/QuadraticProgram/QuadraticProgram.jl
```diff
@@ -356,10 +355,10 @@ function DiffOpt.forward_differentiate!(model::Model)
     partial_grads = if norm(Q) ≈ 0
         -IterativeSolvers.lsqr(LHS', RHS)
```
blegat, line 356: This could also go through LinearSolve: https://github.com/SciML/LinearSolve.jl/blob/f2185983cfd4b25703c498975eb0174174bcc9e5/src/iterative_wrappers.jl#L55

frapac:
As a user of DiffOpt, I confess I am not sure it's a good idea to integrate LinearSolve.jl (compared to, let's say, `ldiv!`). My main concern is the number of dependencies that come with LinearSolve.jl: 
https://github.com/SciML/LinearSolve.jl/blob/f2185983cfd4b25703c498975eb0174174bcc9e5/Project.toml

matbesancon:
that's a fair point, we could readjust the design to just have an abstract attribute on top of which people can choose what solver to take

odow:
Yes, 126 dependencies seems like a lot... https://juliahub.com/ui/Packages/LinearSolve/WR6RC/1.35.0?page=1

matbesancon:
following the discussion, I changed the design to offer an abstract interface, people can change the linear system solve without requiring LinearSolve, since we do not batch, use cache or anything

src/QuadraticProgram/QuadraticProgram.jl
```diff
@@ -395,4 +396,22 @@ function DiffOpt._get_dA(model::Model, ci::LE)
...
+struct LinearAlgebraSolver <: MOI.AbstractOptimizerAttribute end
+
+solve_system(::Nothing, LHS, RHS) = LHS \ RHS
+# See https://github.com/JuliaLang/julia/issues/32668
+solve_system(::Nothing, LHS, RHS::SparseVector) = LHS \ Vector(RHS)
+
+MOI.supports(::Model, ::LinearAlgebraSolver) = true
+MOI.get(model::Model, ::LinearAlgebraSolver) = model.linear_solver
+function MOI.set(model::Model, ::LinearAlgebraSolver, linear_solver)
+    model.model_constructor = model_constructor
```
blegat, line 414:
```suggestion
    model.linear_solver = linear_solver
```

blegat:
Looks good, will be useful for the rank-1 example. Could you add a test for setting and getting the attribute ?

matbesancon:
just did, should be good once CI passes

matbesancon:
I also added an `iterative` argument, which the linear solver decides to handle (the default boils down to lsqr or `\`)

test/solver_interface.jl
```diff
@@ -32,3 +32,21 @@ end
...
+    IterativeSolvers.lsqr(LHS, RHS)
+end
+
+@testset "Setting the linear solver in the quadratic solver" begin
+    model = DiffOpt.QuadraticProgram.Model()
+    @test MOI.supports(model, DiffOpt.QuadraticProgram.LinearAlgebraSolver())
+    @test MOI.get(model, DiffOpt.QuadraticProgram.LinearAlgebraSolver()) === nothing
+    MOI.set(model, DiffOpt.QuadraticProgram.LinearAlgebraSolver(), TestSolver())
+    @test MOI.get(model, DiffOpt.QuadraticProgram.LinearAlgebraSolver()) !== nothing
+    MOI.empty!(model)
```
blegat, line 50: If it is an optimizer attribute, it should not be reset by `MOI.empty!`

matbesancon: fixed

test/solver_interface.jl
```diff
@@ -32,3 +32,21 @@ end
...
+    @test MOI.get(model, DiffOpt.QuadraticProgram.LinearAlgebraSolver()) !== nothing
```
matbesancon, line 51:
```suggestion
    @test MOI.get(model, DiffOpt.QuadraticProgram.LinearAlgebraSolver()) == TestSolver()
```

blegat (approved):

PR #231, opened 2023-02-27
Add conic to ridge sentitivity

blegat:
Preview at https://jump.dev/DiffOpt.jl/previews/PR231

I guess it won't work because we need to implement the transformation of diff attributes through the QuadtoSOC bridge. The nice thing is that if we implement that attribute then we can do conic program with quadratic objective.
The users can always formulate it as a conic program but that would give a way for the user to get the sensitivities in the quadratic objective format directly.
The transformations of the diff attributes are not so easy though since it's not an affine transformation but I can take a look, an argument similar to the dual of QuadtoSOC could go through

Requires

- [x] https://github.com/jump-dev/DiffOpt.jl/pull/232
- [x] https://github.com/matbesancon/MathOptSetDistances.jl/pull/70
- [x] https://github.com/jump-dev/MathOptInterface.jl/pull/2115


matbesancon:
@blegat what is left here now that the other PR is merged?

blegat:
Turned out to be a rabbit hole :laughing:

matbesancon:
From the errors, all the compat entries linked to tests should go in test/Project.toml instead

src/copy_dual.jl
```diff
@@ -0,0 +1,182 @@
+struct ObjectiveFunctionAttribute{A,F}
```
matbesancon, line 1: Docstring for this one?

matbesancon (approved):

PR #232, opened 2023-03-01
Implement ForwardConstraintFunction for QuadToSOCBridge

blegat:


src/bridges.jl
```diff
@@ -30,3 +30,89 @@ function MOI.get(
...
+        end
+    end
+    MOI.set(model, attr, bridge.soc, diff_aff)
+    return
+end
+
+"""
+    dU_from_dQ!(dQ, U)
+
+Overwrite `dQ` into `dU` the solution of
```
matbesancon, line 80: the equation is not super clear here, I cannot deduce whether `dQ` is just an in-place output matrix?
In that case this line below
`dd = view(U, 1:i, j)'view(dQ, 1:i, i)`
is strange since it uses the initial value of dQ

src/bridges.jl
```diff
@@ -30,3 +30,89 @@ function MOI.get(
...
+                dd = view(U, 1:i, j)'view(dQ, 1:i, i)
+                dQ[i, j] -= dd
+            elseif i == j
+                dQ[i, j] /= 2
+            else
+                dQ[i, j] = 0
+            end
+        end
+        Ut = LinearAlgebra.UpperTriangular(view(U, 1:j, 1:j))'
+        LinearAlgebra.ldiv!(Ut, view(dQ, 1:j, j))
```
matbesancon, line 115: we modify in place Ut that we do not return?

src/bridges.jl
matbesancon: Does the function also modify U? If so, it should be specified somewhere

matbesancon:
the test file is also not included yet in test/runtests.jl

src/bridges.jl
blegat: In `LinearAlgebra.ldiv!`, it's actually the second argument that is modified. It's a bit confusing as `!` usually modifies the first one.

src/utils.jl
```diff
@@ -62,3 +75,62 @@ Base.getindex(::IdentityMap, index) = index
 function sparse_array_representation(func::MOI.AbstractFunction, num_variables)
     return sparse_array_representation(func, num_variables, IdentityMap())
 end
+
+struct SparseVectorAffineFunction{T}
```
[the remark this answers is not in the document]
blegat: Done

src/bridges.jl
blegat: > In that case this line below
> `dd = view(U, 1:i, j)'view(dQ, 1:i, i)`
> is strange since it uses the initial value of dQ

It uses the part of `dQ` that was already modified

blegat: I have clarified it in latest commit

src/bridges.jl
```diff
@@ -30,3 +30,94 @@ function MOI.get(
...
+        end
+    end
+    MOI.set(model, attr, bridge.soc, diff_aff)
+    return
+end
+
+"""
+    dU_from_dQ!(dQ, U)
+
+Overwrite `dQ` into `dU` the solution of
```
matbesancon, line 80: > Overwrite `dQ` into `dU`

sounds like we have a dangling argument dU that is not present.

blegat: Is it clearer now ?

src/utils.jl
```diff
@@ -62,3 +75,73 @@ Base.getindex(::IdentityMap, index) = index
...
+        constants::Vector{T}
+    end
+
+The vector-valued affine function ``A x + b``, where:
+
+* ``A`` is the sparse matrix given by `terms`
+* ``b`` is the vector `constants`
+"""
+struct SparseVectorAffineFunction{T} <: MOI.AbstractVectorFunction
+    terms::SparseMatrixCSC{T,Int64}
```
matbesancon, line 91: Int should be sufficient everywhere no, since these are indices

```suggestion
        terms::SparseMatrixCSC{T,Int}
        constants::Vector{T}
    end

The vector-valued affine function ``A x + b``, where:

* ``A`` is the sparse matrix given by `terms`
* ``b`` is the vector `constants`
"""
struct SparseVectorAffineFunction{T} <: MOI.AbstractVectorFunction
    terms::SparseMatrixCSC{T,Int}
```

matbesancon:
yes all good for me!

PR #235, opened 2023-04-03
import as for MOI

blegat:


matbesancon:
is there a minimum Julia version for this, and are we enforcing it?

odow:
Works in Julia 1.6. Maybe even 1.5?

matbesancon:
alright then perfect!

PR #237, opened 2023-04-05
Fix conic objective reverse diff

blegat:
This PR throws the Conic diff through all the QP tests we have with no quadratic parts, this already caught one bug that is fixed in this PR


blegat:
The objective function returned was incorrect for backward differentiation. We should have `πz[end] * g[1:n] - πz[1:n] * g[end]` and we had `πz[end] * g - πz * g[end]` where `n` is the number of variables.

matbesancon (approved):

PR #240, opened 2023-04-05
Add back qp_test

blegat:


matbesancon:
Was something removed or turned into nothing before?

blegat:
Yes in https://github.com/jump-dev/DiffOpt.jl/pull/237

Issue #202, opened 2022-03-18
Test more exotic cases


blegat:
Currently, we only test giving QPs in standard form and use QPDiff or giving CPs in standard for and use ConicDiff.
The user might not give these form and get errors as it's not tested.
We should add tests for this, e.g., adding a constraint in NormOneCone or NormInfinityCone to QPDiff (as it can be rewritten as LP).
This might encouter missing setters and getters for the DiffOpt attributes for a few bridges that we should add. Similar to
https://github.com/jump-dev/DiffOpt.jl/blob/3d7e38530fc0d1251abd4a25c279a82c48556f34/src/bridges.jl#L1-L32

odow:
Closing in favor of https://github.com/jump-dev/DiffOpt.jl/issues/102

Issue #195, opened 2022-03-07
Replace Ipopt by HiGHS


joaquimg:

odow:
Blocked by https://github.com/jump-dev/HiGHS.jl/issues/157

Issue #109, opened 2021-06-21
Add missing tests in the end of: Differentiating conic with PSD and POS constraints


joaquimg:

odow:
x-ref https://github.com/jump-dev/DiffOpt.jl/blob/edb31fe847afa0c96bd1f37ebb8b1c33db08786c/test/conic_program.jl#L574-L578
but there are lots of places in the tests that could be improved.

We could potentially roll this into #102. No need to have lots of issues open saying "add a test."

joaquimg:
replaced by #102 as suggested

Issue #107, opened 2021-06-21
add tests for Differentiating simple SOCP


joaquimg:

odow:
Closing in favor of https://github.com/jump-dev/DiffOpt.jl/issues/102

PR #244, opened 2023-04-11
Convert tests to functional form

### LINKED ISSUE #108: Split moi_wrapper test file in conic and qp

### LINKED ISSUE #110: put tests into functions

odow:
Closes #110 
Closes #108 

Another quite large diff that is just moving things around. This converts all the tests into the function `test_` form to remove global state (some test files were just one large global script!).

I also removed the doc examples from the test script, since these get run and tested when the docs are built: 
https://github.com/jump-dev/DiffOpt.jl/blob/190beb854bb2c37cfd8a42afc8eda99fec095612/docs/make.jl#L33-L37
and not all tutorials were tested anyway.

This let's us remove a bunch of packages from the tests like Flux and Plots, which should help speed things up. 

I haven't addressed #195. That can wait for another PR.

Also, I noticed that there are a _lot_ of comments in the tests with various things commented out and TODOs. I've left for now, but there are lot of open issues related to testing (#202, #109, #108, #107, #106, #102) , so the whole package could use some more work on that front.


odow:
And here's a reason to refactor the tests. 

Tests are failing because it looks like https://github.com/jump-dev/DiffOpt.jl/blob/master/test/sensitivity_index_issue.jl was never actually tested.

matbesancon:
Thanks!

odow:
I never use it, and it can cause confusing bugs: https://github.com/JuliaPolyhedra/CDDLib.jl/pull/82.

But we can add it back...

PR #246, opened 2023-04-12
Split linear program and quad programs

blegat:


test/linear_program.jl
```diff
@@ -0,0 +1,251 @@
+# Copyright (c) 2020: Akshay Sharma and contributors
+#
+# Use of this source code is governed by an MIT-style license that can be found
+# in the LICENSE.md file or at https://opensource.org/licenses/MIT.
+
+module TestLinearProgram
+
+using Test
+import DelimitedFiles
```
odow, line 9: You don't need all of these imports.

PR #247, opened 2023-04-13
Remove Ipopt from tests

### LINKED ISSUE #195: Replace Ipopt by HiGHS

odow:
Closes #195

Let's see if the dual bugs in HiGHS are now fixed.


blegat:
Note that most tests are disabled without https://github.com/jump-dev/DiffOpt.jl/pull/240

joaquimg:
Nowadays Ipopt is needed, so no need to drop some test.

PR #248, opened 2023-04-19
Add nearest correlation example

blegat:


matbesancon:
@blegat just run formatter and good to go I think

PR #253, opened 2023-12-19
Use SlackBridgePrimalDualStart

blegat:
Thanks to https://github.com/jump-dev/MathOptInterface.jl/pull/2194


src/copy_dual.jl
```diff
@@ -173,8 +102,10 @@ function _copy_dual(dest::MOI.ModelLike, src::MOI.ModelLike, index_map)
             MOI.ConstraintDual(),
         )
     end
-    MOI.set(dest, ObjectiveDualStart(), -1.0)
-    return MOI.set(dest, ObjectiveSlackGapPrimalStart(), 0.0)
+    # Same as in `JuMP.set_start_values`
+    # Needed for models which bridge `min f(x)` into `min t; t >= f(x)`.
```
blegat, line 106:
```suggestion
    # Needed for models which bridge `min f(x)` into `min t such that t >= f(x)`.
```

Issue #94, opened 2021-05-13
Diff information of dual variables


matbesancon:
Following a discussion on https://github.com/jump-dev/DiffOpt.jl/pull/91 it might be interesting to add the attributes for dual variables

joaquimg:
The case for NLP is done in #260

PR #260, opened 2024-12-06
🚀 Add NonLinearProgram Support to DiffOpt.jl

andrewrosemberg:
This PR introduces a new module, `NonLinearProgram`, to extend DiffOpt.jl's functionality for differentiating nonlinear optimization problems (NLPs). The implementation integrates with JuMP-based nonlinear models and supports advanced derivative computation through custom evaluator and differentiation logic.

### 🆕 Features

- **Nonlinear Model Differentiation**:
  - Forward and reverse differentiation for user-specified primal variables (`focus_vars`) and dual variables (`focus_duals`) with respect to a given set of parameters.
  - Extensible API for handling NLP-specific sensitivities using custom utilities.

- **Core Structures**:
  - `Cache`: Stores primal variables, parameters, evaluator, and constraints for efficient reuse.
  - `ForwCache`: Holds results of forward differentiation, including sensitivities for specified variables.
  - `ReverseCache`: Holds results of reverse differentiation (implemented in this PR).

- **Integration with DiffOpt API**:
  - Implements `DiffOpt.AbstractModel` for seamless compatibility with DiffOpt's API.
  - Provides `forward_differentiate!` and `reverse_differentiate!` functions for NLPs.

### 🔧 How It Works

1. **Custom Sensitivity Calculations**:
   - Utilizes implicit function differentiation (similar to sIPOPT) to compute derivatives of the optimization problem with respect to selected parameters.
   - Allows differentiation of only specific variables and duals, offering a fine-grained approach compared to existing DiffOpt modules.

2. **Forward Differentiation**:
   - Computes Jacobian products for selected primal and dual variables using the provided parameters.
   - Results are cached for efficient access through the DiffOpt API.

3. **Reverse Differentiation**:
   - Computes transpose Jacobian products for selected primal and dual variables with respect to the parameters.
   - Results are cached in `ReverseCache`.

### 📜 Implementation Highlights

- **Forward Differentiation**:
  ```julia
  DiffOpt.forward_differentiate!(model)
  ```
  - Calculates sensitivity of variables and duals ` w.r.t. `params`.
  - Results are cached in `ForwCache`.

- **Reverse Differentiation**:
  ```julia
  DiffOpt.reverse_differentiate!(model)
  ```
  - Computes transpose Jacobian products for parameters with respect to the primal and dual variables.
  - Results are cached in `ReverseCache`.

- **Custom Utilities**:
  - Leverages `create_evaluator`, `compute_sensitivity`, and other utilities from `nlp_utilities.jl` for efficient derivative computation.

#### Forward Differentiation
```julia
using DiffOpt
using JuMP
using Ipopt

# Define a JuMP model
model = Model(() -> DiffOpt.diff_optimizer(Ipopt.Optimizer))
@variable(model, p ∈ MOI.Parameter(0.1))
@variable(model, x >= p)
@variable(model, y >= 0)
@objective(model, Min, x^2 + y^2)
@constraint(model, con, x + y >= 1)

# Solve
JuMP.optimize!(model)

# set parameter pertubations
MOI.set(model, DiffOpt.ForwardParameter(), params[1], 0.2)

# forward differentiate
DiffOpt.forward_differentiate!(model)

# Retrieve sensitivities
dx = MOI.get(model, DiffOpt.ForwardVariablePrimal(), x)
dy = MOI.get(model, DiffOpt.ForwardVariablePrimal(), y)
```

#### Reverse Differentiation
```julia

# Set Primal Pertubations
MOI.set(model, DiffOpt.ReverseVariablePrimal(), x, 1.0)

# Reverse differentiation
DiffOpt.reverse_differentiate!(model)

# Retrieve reverse sensitivities (example usage)
dp= MOI.get(model, DiffOpt.ReverseParameter(), p)
```

### 🚧 TODO
- [x] **Forward Differentiation**: Implemented and validated for custom primal and dual variables.
- [x] **Reverse Differentiation**: Implemented to compute transpose Jacobian products.
- [x] **Tests**: Add comprehensive test cases for forward and reverse differentiation.
- [x] **Documentation**: Extend user documentation with more examples and edge case scenarios.

### 🛠 Future Work
- Allow different methods to factorize the "M" matrix and solve the linear system of equations needed.


test/data/nlp_problems.jl
```diff
@@ -0,0 +1,345 @@
...
+    else
+        @objective(model, Max, -(1 - x[1])^2 - p3 * (x[2] - x[1]^2)^2)
+    end
+
+    return model, x, [g_1; g_2], [p; p2; p3]
+end
+
+################################################
+#=
+From sIpopt paper: https://optimization-online.org/2011/04/3008/
```
frapac, line 30: very nice!

src/NonLinearProgram/nlp_utilities.jl
```diff
@@ -0,0 +1,556 @@
...
+# Use of this source code is governed by an MIT-style license that can be found
+# in the LICENSE.md file or at https://opensource.org/licenses/MIT.
+
+"""
+    fill_off_diagonal(H)
+
+Filling the off-diagonal elements of a sparse matrix to make it symmetric.
+"""
+function fill_off_diagonal(H)
+    ret = H + H'
```
frapac, line 12: Do we assume `H` is lower-triangular?

src/NonLinearProgram/nlp_utilities.jl
```diff
@@ -0,0 +1,556 @@
...
+#
+# Use of this source code is governed by an MIT-style license that can be found
+# in the LICENSE.md file or at https://opensource.org/licenses/MIT.
+
+"""
+    fill_off_diagonal(H)
+
+Filling the off-diagonal elements of a sparse matrix to make it symmetric.
+"""
+function fill_off_diagonal(H)
```
frapac, line 11: Maybe constrain the input type? `H::SparseMatrixCSC`

src/NonLinearProgram/nlp_utilities.jl
```diff
@@ -0,0 +1,556 @@
...
+
+Compute the optimal Hessian of the Lagrangian.
+"""
+function compute_optimal_hessian(
+    model::Model,
+    rows::Vector{MOI.Nonlinear.ConstraintIndex},
+)
+    sense_multiplier = sense_mult(model)
+    evaluator = model.cache.evaluator
+    y = [model.y[model.model.nlp_index_2_constraint[row].value] for row in rows]
```
frapac, line 36: Is there a proper getter to access `model.model.nlp_index_2_constraint`?

src/NonLinearProgram/nlp_utilities.jl
```diff
@@ -0,0 +1,556 @@
...
+    hessian_sparsity = MOI.hessian_lagrangian_structure(evaluator)
+    I = [i for (i, _) in hessian_sparsity]
+    J = [j for (_, j) in hessian_sparsity]
+    V = zeros(length(hessian_sparsity))
+    # The signals are being sdjusted to match the Ipopt convention (inner.mult_g)
```
frapac, line 41: `sdjusted` => `adjusted`

src/NonLinearProgram/nlp_utilities.jl
```diff
@@ -0,0 +1,556 @@
...
+    V = zeros(length(jacobian_sparsity))
+    MOI.eval_constraint_jacobian(evaluator, V, model.x)
+    A = SparseArrays.sparse(I, J, V, length(rows), length(model.x))
+    return A
+end
+
+"""
+    compute_optimal_hess_jac(evaluator::MOI.Nonlinear.Evaluator, rows::Vector{JuMP.ConstraintRef}, x::Vector{JuMP.VariableRef})
+
+Compute the optimal Hessian of the Lagrangian and Jacobian of the constraints.
```
frapac, line 77: What is the "optimal` Hessian? Do you mean the Hessian evaluated at the solution?

src/NonLinearProgram/nlp_utilities.jl
```diff
@@ -0,0 +1,556 @@
...
+        MOI.ScalarQuadraticFunction{Float64},
+        MOI.ScalarAffineFunction{Float64},
+    },
+    S<:MOI.GreaterThan,
+}
+    return true
+end
+
+"""
+    find_inequealities(cons::Vector{JuMP.ConstraintRef})
```
frapac, line 151: `find_inequealities` => `find_inequalities`

src/NonLinearProgram/nlp_utilities.jl
```diff
@@ -0,0 +1,556 @@
...
+    for con in keys(model.list_of_constraint)
+        if is_less_inequality(con)
+            leq_locations[model.constraints_2_nlp_index[con].value] = true
+        end
+        if is_greater_inequality(con)
+            geq_locations[model.constraints_2_nlp_index[con].value] = true
+        end
+    end
+    return findall(x -> x == 1, leq_locations),
+    findall(x -> x == 1, geq_locations)
```
frapac, line 168: I would put the two `findall`on the same line

src/NonLinearProgram/nlp_utilities.jl
```diff
@@ -0,0 +1,556 @@
...
+        # dual.(con) * (- sense_multiplier)
+        if sense_multiplier == 1.0
+            V_U[num_vars+num_geq+i] <= -tol &&
+                @info "Dual of leq constraint must be positive" i V_U[num_vars+i]
+        else
+            V_U[num_vars+num_geq+i] >= tol &&
+                @info "Dual of leq constraint must be negative" i V_U[num_vars+i]
+        end
+    end
+    return X,
```
frapac, line 268: I would maybe do a pass of JuliaFormatters there

src/NonLinearProgram/nlp_utilities.jl
```diff
@@ -0,0 +1,556 @@
...
+    leq_locations,
+    geq_locations,
+    ineq_locations,
+    vcat(has_up, collect(num_vars+num_geq+1:num_vars+num_geq+num_leq)),
+    vcat(has_low, collect(num_vars+1:num_vars+num_geq)),
+    cons
+end
+
+"""
+    build_M_N(model::Model, cons::Vector{MOI.Nonlinear.ConstraintIndex}, _X::AbstractVector, _V_L::AbstractVector, _X_L::AbstractVector, _V_U::AbstractVector, _X_U::AbstractVector, leq_locations::Vector{Z}, geq_locations::Vector{Z}, ineq_locations::Vector{Z}, has_up::Vector{Z}, has_low::Vector{Z})
```
frapac, line 282: How about renaming this function `build_sensitivity_matrices` ?

src/NonLinearProgram/nlp_utilities.jl
```diff
@@ -0,0 +1,556 @@
...
+    for (i, j) in enumerate(leq_locations)
+        A[j, num_vars+i] = -1
+    end
+    # Partial second derivative of the lagrangian wrt primal solution and parameters
+    ∇ₓₚL = spzeros(num_vars + num_ineq, num_parms)
+    ∇ₓₚL[1:num_vars, :] = hessian[primal_idx, params_idx]
+    # Partial derivative of the equality constraintswith wrt parameters
+    ∇ₚC = jacobian[:, params_idx]
+
+    # M matrix
```
frapac, line 358: Comment is appreciated!

src/NonLinearProgram/nlp_utilities.jl
```diff
@@ -0,0 +1,556 @@
...
+    # M = [
+    #     [W A' -I I];
+    #     [A 0 0 0];
+    #     [V_L 0 (X - X_L) 0]
+    #     [V_U 0 0 0 (X_U - X)]
+    # ]
+    len_w = num_vars + num_ineq
+    M = spzeros(
```
frapac, line 366: I think there should be a cleaner way to build `M` and `N`, with less allocations. Let me try to build a MWE, we can assess later if we should address this in this PR or later.

src/NonLinearProgram/nlp_utilities.jl
```diff
@@ -0,0 +1,556 @@
...
+    N[1:len_w, :] = ∇ₓₚL
+    N[len_w+1:len_w+num_cons, :] = ∇ₚC
+
+    return M, N
+end
+
+"""
+    inertia_corrector_factorization(M::SparseMatrixCSC, num_w, num_cons; st=1e-6, max_corrections=50)
+
+Inertia correction for the factorization of the KKT matrix. Sparse version.
```
frapac, line 399: The theoretical question is: do we need to run inertia correction at the optimum? It could be necessary if SOSC is not satisfied at the solution indeed, but in that case I am not sure the sIpopt's formula remains valid.

src/NonLinearProgram/nlp_utilities.jl
```diff
@@ -0,0 +1,556 @@
...
+        _V_U,
+        _X_U,
+        leq_locations,
+        geq_locations,
+        ineq_locations,
+        has_up,
+        has_low,
+    )
+
+    # Sesitivity of the solution (primal-dual_constraints-dual_bounds) w.r.t. the parameters
```
frapac, line 490: `Sesitivity` => `Sensitivity`

src/NonLinearProgram/nlp_utilities.jl
```diff
@@ -0,0 +1,556 @@
...
+    num_cons = get_num_constraints(model)
+    num_ineq = length(ineq_locations)
+    K = inertia_corrector_factorization(M, num_vars + num_ineq, num_cons) # Factorization
+    if isnothing(K)
+        return zeros(size(M, 1), size(N, 2)), K, N
+    end
+    ∂s = zeros(size(M, 1), size(N, 2))
+    # ∂s = - (K \ N) # Sensitivity
+    ldiv!(∂s, K, N)
+    ∂s = -∂s
```
frapac, line 501: Maybe we should add a comment to explain why we are multiplying by `-1` there

frapac (commented):
good job @andrewrosemberg ! I appreciate the numerous unit-tests you have shipped with this PR. 

I will try to run your code locally on small NLP instances, to assess how fast is the current implementation (my main concern is the total number of allocations). I will try also to test your code on parameterized OPF instances, to assess how far we can get in term of size.

src/NonLinearProgram/nlp_utilities.jl
andrewrosemberg: Not that I know of. @joaquimg , do you ?

src/NonLinearProgram/nlp_utilities.jl
andrewrosemberg: yes, will edit!

src/NonLinearProgram/nlp_utilities.jl
andrewrosemberg: This was made by formatter. I believe we need it to pass the format test

src/NonLinearProgram/nlp_utilities.jl
andrewrosemberg: I did, this was all done by formatter

src/NonLinearProgram/nlp_utilities.jl
andrewrosemberg: I think you are right, but couldn't there be numerical errors? 

Moreover, I believe the rest of DiffOpt just assumes that the necessary conditions hold (even if they don't). The thing is that their method doesn't fail when it doesn't hold, while sIpopt method appears to throw singular matrix error when either the conditions don't hold or when we have numerical errors. Therefore, I tried to make it consistent.

frapac:
@andrewrosemberg Playing with your branch right now, I must say I like DiffOpt's interface! 

I did a dummy mistake when solving the problem HS15 and retrieving the sensitivity. A MWE is: 

```julia
model = Model(() -> DiffOpt.diff_optimizer(Ipopt.Optimizer))
@variable(model, p[1:2] ∈ MOI.Parameter.([100.0, 1.0]))
@variable(model, x[1:2])
set_upper_bound.(x[1], 0.5)
@objective(model, Min, p[1] * (x[2] - x[1]^2)^2 +  (p[2] - x[1])^2)
@constraint(model, x[1] * x[2] >= 1.0)
@constraint(model, x[1] + x[2]^2 >= 0.0)

optimize!(model)
# set parameter pertubations
MOI.set(model, DiffOpt.ForwardParameter(), p[1], 1.0)

# # forward differentiate
DiffOpt.forward_differentiate!(model)

Δx = [
    MOI.get(model, DiffOpt.ForwardVariablePrimal(), var) for
    var in x
]

```

I forgot to specify the sensitivity for `p[2]`, resulting in an error. I am wondering what would be the expected behavior here:
- by default, should we set the sensitivity to `0.0` ? 
- or should we raise a proper error message if the user has forgotten to specify some sensitivities?

src/NonLinearProgram/nlp_utilities.jl
Robbybp: Maybe we should log a warning if inertia correction is necessary (apologies if you're already doing this) or add an option `allow_intertia_correction` that defaults to true?

frapac:
Also, if I query the solution after calling `forward_differentiate` I get an error. Is this expected? 

The MWE is: 
```julia
model = Model(() -> DiffOpt.diff_optimizer(Ipopt.Optimizer))
@variable(model, p[1:2] ∈ MOI.Parameter.([100.0, 1.0]))
@variable(model, x[1:2])
set_upper_bound.(x[1], 0.5)
@objective(model, Min, p[1] * (x[2] - x[1]^2)^2 +  (p[2] - x[1])^2)
@constraint(model, x[1] * x[2] >= 1.0)
@constraint(model, x[1] + x[2]^2 >= 0.0)
optimize!(model)
# set parameter pertubations
MOI.set(model, DiffOpt.ForwardParameter(), p[1], 1.0)
MOI.set(model, DiffOpt.ForwardParameter(), p[2], 1.0)

# # forward differentiate
DiffOpt.forward_differentiate!(model)

JuMP.value.(x)

```

Output: 
```
┌ Warning: The model has been modified since the last call to `optimize!` (or `optimize!` has not been called yet). If you are iteratively querying solution information and modifying a model, query all the results first, then modify the model.
└ @ JuMP ~/.julia/packages/JuMP/CU7H5/src/optimizer_interface.jl:1085
ERROR: LoadError: OptimizeNotCalled()
```

andrewrosemberg:
> I forgot to specify the sensitivity for `p[2]`, resulting in an error. I am wondering what would be the expected behavior here:
> 
> * by default, should we set the sensitivity to `0.0` ?
> * or should we raise a proper error message if the user has forgotten to specify some sensitivities?

@frapac good question! I am fine either way. @joaquimg what would be the consistent approach given the rest of DIffOpt?

andrewrosemberg:
> Also, if I query the solution after calling forward_differentiate I get an error. Is this expected?

I don't think this is the desired outcome haha Not sure how to ask MOI to ignore the forward attributes, but I will look into it.  @joaquimg do you have an idea on how to avoid this?

frapac:
@andrewrosemberg Another issue I noted: if we are using non-standard indexing in JuMP (`DenseAxisArray` instead of `Vector{VariableRef}`) I got an error when retrieving the sensitivity 

```
ERROR: LoadError: MethodError: no method matching similar(::Type{Vector{Float64}}, ::Tuple{UnitRange{Int64}})
The function `similar` exists, but no method is defined for this combination of argument types.
```

A MWE is:

```julia
model = Model(() -> DiffOpt.diff_optimizer(Ipopt.Optimizer))
@variable(model, p[1:2] ∈ MOI.Parameter.([100.0, 1.0]))
# N.B: use non-standard indexing
@variable(model, x[0:2])
set_upper_bound.(x[1], 0.5)
@objective(model, Min, p[1] * (x[2] - x[1]^2)^2 +  (p[2] - x[1])^2)
@constraint(model, x[1] * x[2] >= 1.0)
@constraint(model, x[1] + x[2]^2 >= 0.0)
optimize!(model)
# set parameter pertubations
MOI.set(model, DiffOpt.ForwardParameter(), p[1], 1.0)
MOI.set(model, DiffOpt.ForwardParameter(), p[2], 1.0)

# # forward differentiate
DiffOpt.forward_differentiate!(model)

Δx = [
    MOI.get(model, DiffOpt.ForwardVariablePrimal(), var) for
    var in x
]


```

frapac:
Pushing it one step further, I have tried to differentiate an ACOPF instance using DiffOpt. I re-used the code in rosetta-opf, and ported it to DiffOpt. You can find a gist [here](https://gist.github.com/frapac/2e81f6c61661effdce0d8b78f52995b0). 

Two observations: 
- on `case9`, DiffOpt is using inertia correction. This is not expected, as I am sure we satisfy all the good constraints qualifications in `case9`. I am not sure to understand why UMFPACK fails to factorize the KKT system. 
- on `case1354pegase`, I get an out-of-memory error. This is caused by [this line](https://github.com/andrewrosemberg/DiffOpt.jl/blob/ar/NonLinearProgram/src/NonLinearProgram/nlp_utilities.jl#L415), which allocates a dense matrix.

src/NonLinearProgram/nlp_utilities.jl
andrewrosemberg: done!

andrewrosemberg:
> @andrewrosemberg Another issue I noted: if we are using non-standard indexing in JuMP (`DenseAxisArray` instead of `Vector{VariableRef}`) I got an error when retrieving the sensitivity
> 
> ```
> ERROR: LoadError: MethodError: no method matching similar(::Type{Vector{Float64}}, ::Tuple{UnitRange{Int64}})
> The function `similar` exists, but no method is defined for this combination of argument types.
> ```
> 
> A MWE is:

> Δx = [
>     MOI.get(model, DiffOpt.ForwardVariablePrimal(), var) for
>     var in x
> ]
> ```

I believe that is a Julia thing. If you try the same with `value(var)` it will give a similar error. if you iterate by index it works.

src/NonLinearProgram/nlp_utilities.jl
andrewrosemberg: I was following Oscar's steps from here: 

https://github.com/jump-dev/JuMP.jl/blob/54c1239ee964b13012a4553b4e36bfb5579e6aff/docs/src/tutorials/nonlinear/querying_hessians.jl#L204-L221

src/NonLinearProgram/nlp_utilities.jl
andrewrosemberg: done!

andrewrosemberg:
> I forgot to specify the sensitivity for `p[2]`, resulting in an error. I am wondering what would be the expected behavior here:
> 
> * by default, should we set the sensitivity to `0.0` ?
> * or should we raise a proper error message if the user has forgotten to specify some sensitivities?

@frapac I have just set the sensitivities to zero since it seems to be the way that we went in the parametric interface Joaquim did.

andrewrosemberg:
> * on `case9`, DiffOpt is using inertia correction. This is not expected, as I am sure we satisfy all the good constraints qualifications in `case9`. I am not sure to understand why UMFPACK fails to factorize the KKT system.

@frapac , this happens because of the following constraints:

```
JuMP.@constraint(model, branch["angmin"] <= va_fr - va_to <= branch["angmax"])
```

This creates two linearly dependent constraints. Ipopt and sipopt avoid any problems by eliminating the non-active constraint. Currently, our method uses the full KKT. This is very suboptimal but changing requires some careful steps. I will create an Issue to record this but I believe it should be solved in a subsequent PR.

andrewrosemberg:
> * on `case1354pegase`, I get an out-of-memory error. This is caused by [this line](https://github.com/andrewrosemberg/DiffOpt.jl/blob/ar/NonLinearProgram/src/NonLinearProgram/nlp_utilities.jl#L415), which allocates a dense matrix. I would suggest replacing `diagm` by `SparseArrays.spdiagm`

@frapac this is now fixed, thank you :)

docs/src/manual.md
```diff
@@ -16,6 +12,12 @@ For `QuadraticProgram` backend, the package supports following `Function-in-Set`
 |    `ScalarAffineFunction`    |    `GreaterThan`    |
 |    `ScalarAffineFunction`    |    `LessThan`    |
 |    `ScalarAffineFunction`    |    `EqualTo`    |
+|    `ScalarQuadraticFunction`    |    `GreaterThan`    |
+|    `ScalarQuadraticFunction`    |    `LessThan`    |
+|    `ScalarQuadraticFunction`    |    `EqualTo`    |
+|    `ScalarNonlinearFunction`    |    `GreaterThan`    |
+|    `ScalarNonlinearFunction`    |    `LessThan`    |
+|    `ScalarNonlinearFunction`    |    `EqualTo`    |
```
joaquimg, line 20: this section is about the QP backend, we need a separate section for the NLP backend

docs/src/manual.md
```diff
@@ -24,6 +26,7 @@ and the following objective types:
 |   `VariableIndex`   |
 |   `ScalarAffineFunction`   |
 | `ScalarQuadraticFunction`  | 
+| `ScalarNonlinearFunction`  |
```
joaquimg, line 29: same as above

docs/src/usage.md
```diff
@@ -56,3 +56,65 @@ MOI.set(model, DiffOpt.ForwardObjectiveFunction(), ones(2) ⋅ x)
...
+model = Model(() -> DiffOpt.diff_optimizer(Ipopt.Optimizer))
+set_silent(model)
+
+p_val = 4.0
+pc_val = 2.0
+@variable(model, x)
+@variable(model, p in Parameter(p_val))
+@variable(model, pc in Parameter(pc_val))
+@constraint(model, cons, pc * x >= 3 * p)
+@objective(model, Min, 2x)
```
joaquimg, line 74: lets do somethinf slightly nl like x^4

docs/src/usage.md
```diff
@@ -56,3 +56,65 @@ MOI.set(model, DiffOpt.ForwardObjectiveFunction(), ones(2) ⋅ x)
 DiffOpt.forward_differentiate!(model)
 grad_x = MOI.get.(model, DiffOpt.ForwardVariablePrimal(), x)
 ```
+
+3. To differentiate a general nonlinear program, have to use the API for Parameterized JuMP models. For example, consider the following nonlinear program:
```
joaquimg, line 60: In forward mode,
what happens in the NLP backend if one tries to Set something other than parameters?

In reverse mode, 
what happens in the NLP backend if one tries to Get something other than parameters?

Are these cases tested?

src/parameters.jl
```diff
@@ -319,6 +315,16 @@ function MOI.set(
...
+function MOI.set(
+    model::Optimizer,
+    ::ForwardConstraintSet,
+    ci::MOI.ConstraintIndex{MOI.VariableIndex,MOI.Parameter{T}},
+    set::MOI.Parameter,
+) where {T}
+    model.input_cache.dp[ci] = set.value
+    return
+end
+
```
joaquimg, line 327: doesnt it exist already?
https://github.com/jump-dev/DiffOpt.jl/blob/95927fae094da6603cfe02f608d7bf92898bb70d/src/parameters.jl#L307

src/diff_opt.jl
```diff
@@ -135,6 +139,37 @@ MOI.set(model, DiffOpt.ReverseVariablePrimal(), x)
...
+
+A `MOI.AbstractConstraintAttribute` to get output data from forward differentiation for the dual variable.
+
+For instance, to get the sensitivity of the dual of constraint of index `ci` with respect to the parameter perturbation, do the following:
+
+```julia
+MOI.get(model, DiffOpt.ForwardConstraintDual(), ci)
+```
+"""
+struct ForwardConstraintDual <: MOI.AbstractConstraintAttribute end
```
joaquimg, line 169: creat issue to allow using this in other modes

src/diff_opt.jl
```diff
@@ -29,14 +31,16 @@ end
...
+    empty!(cache.dy)
     empty!(cache.scalar_constraints)
     empty!(cache.vector_constraints)
     cache.objective = nothing
     return
 end
 
 """
-    reverse_differentiate!(model::MOI.ModelLike)
+    reverse_differentiate!(model::MOI.ModelLike; kwargs...)
```
joaquimg, line 43: docs for kwargs?

src/diff_opt.jl
```diff
@@ -13,6 +13,8 @@ const MOIDD = MOI.Utilities.DoubleDicts
 
 Base.@kwdef mutable struct InputCache
     dx::Dict{MOI.VariableIndex,Float64} = Dict{MOI.VariableIndex,Float64}()# dz for QP
+    dp::Dict{MOI.ConstraintIndex,Float64} = Dict{MOI.ConstraintIndex,Float64}()
+    dy::Dict{MOI.ConstraintIndex,Float64} = Dict{MOI.ConstraintIndex,Float64}()
```
joaquimg, line 17: add a comment here mentioning it currently only works for NLP

src/NonLinearProgram/nlp_utilities.jl
joaquimg: if there was, it would be a MOI.get...

src/NonLinearProgram/nlp_utilities.jl
```diff
@@ -0,0 +1,585 @@
...
+#
+# Use of this source code is governed by an MIT-style license that can be found
+# in the LICENSE.md file or at https://opensource.org/licenses/MIT.
+
+"""
+    fill_off_diagonal(H)
+
+Filling the off-diagonal elements of a sparse matrix to make it symmetric.
+"""
+function fill_off_diagonal(H::SparseMatrixCSC)
```
joaquimg, line 11: we need underscores in all non-user facing API functions, e.g. `_fill_off_diagonal`

src/NonLinearProgram/nlp_utilities.jl
```diff
@@ -0,0 +1,585 @@
...
+        end
+    end
+    # value and dual of the upper bounds
+    V_U = spzeros(num_vars + num_ineq)
+    X_U = spzeros(num_vars + num_ineq)
+    for (i, j) in enumerate(has_up)
+        V_U[j] =
+            model.y[form.constraint_upper_bounds[primal_vars[j].value].value] *
+            (-sense_multiplier)
+        # dual.(UpperBoundRef(primal_vars[j])) * (- sense_multiplier)
```
joaquimg, line 242: remove?

src/NonLinearProgram/nlp_utilities.jl
```diff
@@ -0,0 +1,585 @@
...
+    X = [model.x[primal_idx]; s_geq; s_leq]
+
+    # value and dual of the lower bounds
+    V_L = spzeros(num_vars + num_ineq)
+    X_L = spzeros(num_vars + num_ineq)
+    for (i, j) in enumerate(has_low)
+        V_L[j] =
+            model.y[form.constraint_lower_bounds[primal_vars[j].value].value] *
+            sense_multiplier
+        #dual.(LowerBoundRef(primal_vars[j])) * sense_multiplier
```
joaquimg, line 209: remove?

src/NonLinearProgram/nlp_utilities.jl
```diff
@@ -0,0 +1,585 @@
...
+        #
+        X_U[j] = form.upper_bounds[primal_vars[j].value]
+    end
+    for (i, con) in enumerate(cons[leq_locations])
+        # By convention jump dual will allways be negative for leq constraints
+        # but for ipopt it will be positive if min problem and negative if max problem
+        V_U[num_vars+num_geq+i] =
+            model.y[form.nlp_index_2_constraint[con].value] *
+            (-sense_multiplier)
+        # dual.(con) * (- sense_multiplier)
```
joaquimg, line 259: remove?

src/NonLinearProgram/nlp_utilities.jl
joaquimg: inertial correction should be set as a MOI attribute, not as a keyword

src/NonLinearProgram/NonLinearProgram.jl
```diff
@@ -0,0 +1,577 @@
...
+        MOI.LessThan{Float64},
+        # MOI.Interval{Float64},
+        MOI.EqualTo{Float64},
+        MOI.Parameter{Float64},
+    },
+}
+    return true
+end
+
+function add_leq_geq(form::Form, idx::MOI.ConstraintIndex, set::MOI.GreaterThan)
```
joaquimg, line 128: underscores...

src/diff_opt.jl
joaquimg: probably should be MOI attributes set on the model, not kwargs

src/jump_moi_overloads.jl
```diff
@@ -337,12 +347,12 @@ end
 
 # JuMP
 
-function reverse_differentiate!(model::JuMP.Model)
-    return reverse_differentiate!(JuMP.backend(model))
+function reverse_differentiate!(model::JuMP.Model; kwargs...)
```
joaquimg, line 350: avoid kwargs if we can use MOI attributes

joaquimg (changes requested):
First pass

joaquimg:
Also, I dont know why codecov is not showing up...

docs/src/usage.md
andrewrosemberg: done!

docs/src/manual.md
andrewrosemberg: Good catch. fixed!

docs/src/manual.md
andrewrosemberg: Good catch. fixed!

docs/src/usage.md
andrewrosemberg: Good point, I have now extended test from Parameter interface for the non POI version.

src/diff_opt.jl
andrewrosemberg: done!

src/parameters.jl
joaquimg: Oh, fine!

src/NonLinearProgram/nlp_utilities.jl
andrewrosemberg: done

src/NonLinearProgram/nlp_utilities.jl
andrewrosemberg: done

src/NonLinearProgram/NonLinearProgram.jl
andrewrosemberg: changed!

src/jump_moi_overloads.jl
andrewrosemberg: changed to MOI.AbstractModelAttribute

src/NonLinearProgram/NonLinearProgram.jl
```diff
@@ -0,0 +1,566 @@
...
+all_params(model::Model) = all_params(model.model)
+all_primal_vars(form::Form) = setdiff(all_variables(form), all_params(form))
+all_primal_vars(model::Model) = all_primal_vars(model.model)
+
+get_num_constraints(form::Form) = length(form.constraints_2_nlp_index)
+get_num_constraints(model::Model) = get_num_constraints(model.model)
+get_num_primal_vars(form::Form) = length(all_primal_vars(form))
+get_num_primal_vars(model::Model) = get_num_primal_vars(model.model)
+get_num_params(form::Form) = length(all_params(form))
+get_num_params(model::Model) = get_num_params(model.model)
```
joaquimg, line 371: all these might need underscores

src/NonLinearProgram/nlp_utilities.jl
joaquimg: underscores?

src/NonLinearProgram/nlp_utilities.jl
```diff
@@ -0,0 +1,501 @@
...
+    end
+    return ret
+end
+
+"""
+    compute_optimal_hessian(evaluator::MOI.Nonlinear.Evaluator, rows::Vector{JuMP.ConstraintRef}, x::Vector{JuMP.VariableRef})
+
+Compute the Hessian of the Lagrangian calculated at the optimal solution.
+"""
+function compute_optimal_hessian(
```
joaquimg, line 30: all these will also need underscores

src/NonLinearProgram/nlp_utilities.jl
```diff
@@ -0,0 +1,501 @@
+# Copyright (c) 2025: Andrew Rosemberg and contributors
+#
+# Use of this source code is governed by an MIT-style license that can be found
+# in the LICENSE.md file or at https://opensource.org/licenses/MIT.
```
joaquimg, line 4: the header should mention the tutorial you got code from

src/NonLinearProgram/nlp_utilities.jl
joaquimg: I think @frapac means, add a better comment referencing the paper you got this from?

src/diff_opt.jl
```diff
@@ -11,8 +11,55 @@
 
 const MOIDD = MOI.Utilities.DoubleDicts
 
+"""
+    LuFactorizationWithInertiaCorrection{T<:Real}
+
+A callable struct to store the parameters for the inertia correction in the
+Lu-factorization. If no inertia correction is needed, it only performs the LU
+factorization.
```
joaquimg, line 19: needs comments about the parameters.

we need to think a bit more about this.

Initially I thought about having the parameters st and max corrections as atrrbitues.

Do we expect people will pass their factorization routines?

src/diff_opt.jl
joaquimg: we need to remove the kwargs

src/jump_moi_overloads.jl
joaquimg: remove the kwargs now?

joaquimg (changes requested):

src/NonLinearProgram/nlp_utilities.jl
andrewrosemberg: done

src/NonLinearProgram/NonLinearProgram.jl
andrewrosemberg: done

src/diff_opt.jl
andrewrosemberg: This allows user to use better factorization routines like the ones in HSL.

src/diff_opt.jl
joaquimg: https://github.com/jump-dev/DiffOpt.jl/issues/94

PR #281, opened 2025-02-22
Add JuMP-like API

### LINKED ISSUE #271: Add a JuMP-like syntax to diff wrt parameters
Follow-up of #262

joaquimg:
closes #271

### DIFF
--- a/src/jump_wrapper.jl
+++ b/src/jump_wrapper.jl
@@ -1,0 +1,63 @@
+"""
+"""
+function diff_model(
+    optimizer_constructor;
+    method = nothing,
+    with_parametric_opt_interface::Bool = false,
+    with_bridge_type = Float64,
+    with_cache::Bool = true,
+)
+    inner = diff_optimizer(
+        optimizer_constructor;
+        method = method,
+        with_parametric_opt_interface = with_parametric_opt_interface,
+        with_bridge_type = with_bridge_type,
+        with_cache = with_cache,
+    )
+    return JuMP.direct_model(inner)
+end
+
+# nonlinear_diff_model
+# conic_diff_model
+# quadratic_diff_model
+
+"""
+"""
+function set_forward_parameter(
+    model::JuMP.Model,
+    variable::JuMP.VariableRef,
+    value::Number,
+)
+    return MOI.set(
+        model,
+        DiffOpt.ForwardConstraintSet(),
+        ParameterRef(variable),
+        value,
+    )
+end
+
+"""
+"""
+function get_reverse_parameter(model::JuMP.Model, variable::JuMP.VariableRef)
+    return MOI.get(
+        model,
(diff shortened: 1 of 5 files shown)


src/jump_wrapper.jl
```diff
@@ -0,0 +1,141 @@
...
+See also: [`conic_diff_model`](@ref), [`quadratic_diff_model`](@ref), [`diff_model`](@ref).
+"""
+function nonlinear_diff_model(
+    optimizer_constructor;
+    with_bridge_type = Float64,
+    with_cache::Bool = true,
+)
+    inner = diff_optimizer(
+        optimizer_constructor;
+        with_parametric_opt_interface = false,
```
blegat, line 40: Add comment on why this is false

test/jump_wrapper.jl
```diff
@@ -0,0 +1,124 @@
...
+end
+
+function test_jump_api()
+    for (MODEL, SOLVER) in [
+            (DiffOpt.diff_model, Ipopt.Optimizer),
+            (DiffOpt.quadratic_diff_model, HiGHS.Optimizer),
+            (DiffOpt.quadratic_diff_model, SCS.Optimizer),
+            (DiffOpt.quadratic_diff_model, Ipopt.Optimizer),
+            # (DiffOpt.conic_diff_model, HiGHS.Optimizer),
+            # (DiffOpt.conic_diff_model, SCS.Optimizer), # conicmodel has a issue with sign
```
blegat, line 37: We need at least one of the three to be enabled otherwise it's not tested

src/jump_wrapper.jl
joaquimg: Yup, I might re rewrite some of the internals to try to ger rid of this

test/jump_wrapper.jl
joaquimg: Yup,
But there is a bug.
Conic returns wrong sign in some cases.

src/NonLinearProgram/NonLinearProgram.jl
```diff
@@ -595,6 +607,7 @@ function MOI.get(
     ci::MOI.ConstraintIndex,
 )
     try
+        # TODO check ci.value's
         idx = model.cache.dual_mapping[ci.value]
         return model.forw_grad_cache.dual_Δs[idx]
     catch
```
joaquimg, line 613: @andrewrosemberg , this way of getting the map does not seem to match perfectly with the one below

andrewrosemberg: My logic was:
 - `dual_Δs` is indexed by NLP model constraint index, so we need to go from MOI to NLP using `dual_mapping`
 - In the one below `Δp` is indexed my NLP Parameter index, so we need to go from MOI variable index to NLP parameter index, thus the use of `var2param`

Issue #283, opened 2025-03-22
No Clear API for `ReverseConstraintFunction` for Conic Constraints


andrewrosemberg:
The following errors and it is not clear how to correctly fetch the `ReverseConstraintFunction` perturbations for conic constraints:

```julia
model = JuMP.Model(() -> DiffOpt.diff_optimizer(SCS.Optimizer))
set_silent(model)

x = @variable(model)
y = @variable(model)
t = @variable(model)

ceq = @constraint(model, -1.0t == -1.0)

cnon = @constraint(model, 1.0y >= 1 / √2)
csoc = @constraint(model, [1.0t, 1.0x, 1.0y] in MOI.SecondOrderCone(3))

@objective(model, Min, 1.0x)

optimize!(model)

MOI.set(
      model,
      DiffOpt.ReverseVariablePrimal(),
      y,
      1.0
  )

  DiffOpt.reverse_differentiate!(model)

JuMP.coefficient(MOI.get(model, DiffOpt.ReverseConstraintFunction(), cnon), y)
```

Errors with:

```
ERROR: MethodError: no method matching coefficient(::DiffOpt.MatrixVectorAffineFunction{…}, ::MathOptInterface.VariableIndex)

Closest candidates are:
  coefficient(::DiffOpt.MatrixScalarQuadraticFunction, ::MathOptInterface.VariableIndex)
   @ DiffOpt ~/Workspace/DiffOpt.jl/src/utils.jl:228
  coefficient(::DiffOpt.MatrixScalarQuadraticFunction, ::MathOptInterface.VariableIndex, ::MathOptInterface.VariableIndex)
   @ DiffOpt ~/Workspace/DiffOpt.jl/src/jump_moi_overloads.jl:198
  coefficient(::GenericQuadExpr{C, V}, ::V, ::V) where {C, V}
   @ JuMP ~/.julia/packages/JuMP/xlp0s/src/quad_expr.jl:220
  ...

Stacktrace:
 [1] coefficient(func::DiffOpt.IndexMappedFunction{…}, vi::MathOptInterface.VariableIndex)
   @ DiffOpt ~/Workspace/DiffOpt.jl/src/jump_moi_overloads.jl:247
 [2] coefficient(func::DiffOpt.IndexMappedFunction{DiffOpt.IndexMappedFunction{…}}, vi::MathOptInterface.VariableIndex)
[...]

PR #291, opened 2025-05-20
Fix standard_form

joaquimg:
@blegat is this (part of) the solution for #283 ?

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -263,7 +263,9 @@ function Base.convert(
         func.constants,
     )
 end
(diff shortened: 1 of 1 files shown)


joaquimg:
`JuMP.coefficient` does not go well with vector affine functions because in the JuMP level they are a `Vector` of `AffExpr`, hence there is no `JuMP.coefficient(::SomeFunction, ::Integer, ::VariableRef)`.
Possibly the more JuMP-like API is allowing getindex to add a lazy layer on top of `MatrixVectorAffineFunction`, but that might be too complex to maintain.

PR #294, opened 2025-07-07
Update jump-api pr

joaquimg:


src/NonLinearProgram/NonLinearProgram.jl
```diff
@@ -601,6 +607,7 @@ function MOI.get(
     ci::MOI.ConstraintIndex,
 )
     try
+        # TODO check ci.value's
         idx = model.cache.dual_mapping[ci.value]
         return model.forw_grad_cache.dual_Δs[idx]
     catch
```
joaquimg, line 613: @andrewrosemberg, this way of getting the map does not seem to match perfectly with the one below

PR #282, opened 2025-02-22
Implement dual of parameter anywhere

andrewrosemberg:
It allows for the dual of parameters anywhere in the model.

```julia
using JuMP, DiffOpt, HiGHS

model = Model(() -> DiffOpt.diff_optimizer(Ipopt.Optimizer))
set_silent(model)

p_val = 4.0
pc_val = 2.0
@variable(model, x)
@variable(model, p in Parameter(p_val))
@variable(model, pc in Parameter(pc_val))
@constraint(model, cons, pc * x >= 3 * p)
@objective(model, Min, x^4)
optimize!(model)

direction_p = 3.0
MOI.set(model, DiffOpt.ForwardConstraintSet(), ParameterRef(p), Parameter(direction_p))
DiffOpt.forward_differentiate!(model)

dual(p) # works

dual(pc) # also works
```

Issue #297, opened 2025-08-15
Improve parameter checks or allow mixing parameters and functions when setting sensitivities


andrewrosemberg:
**Summary**
When setting sensitivities (sensibs) the current API inconsistently handles mixing parameters and functions. Need to either enforce clearer validation with better errors or support the mixed usage.

**Decision Needed**
Do we (A) strictly enforce separation with clear errors, or (B) support mixing? This issue covers implementing whichever path is chosen (may split later if needed).

TODO from: https://github.com/jump-dev/DiffOpt.jl/pull/281

PR #303, opened 2025-08-20
Objective Sensitivity 2.0

andrewrosemberg:
Had issues with my old pr (https://github.com/jump-dev/DiffOpt.jl/pull/282). So this one is the updated version.


docs/src/usage.md
```diff
@@ -117,4 +117,71 @@ DiffOpt.reverse_differentiate!(model)
...
+@variable(model, p in Parameter(p_val))
+@variable(model, pc in Parameter(pc_val))
+@constraint(model, cons, pc * x >= 3 * p)
+@objective(model, Min, x^4)
+optimize!(model)
+
+direction_p = 3.0
+MOI.set(model, DiffOpt.ForwardConstraintSet(), ParameterRef(p), Parameter(direction_p))
+DiffOpt.forward_differentiate!(model)
+
```
blegat: Remove empty line

docs/src/usage.md
```diff
@@ -117,4 +117,71 @@ DiffOpt.reverse_differentiate!(model)
...
+```
+
+Using Lagrangian duality we could already calculate the objective sensitivity with respect to parameters that appear in the RHS of the constraints (e.g, `cons` in this case for parameter `p`) - i.e. The objective sensitivity w.r.t. a parameter change in the RHS of the constraints is given by the optimal multiplier.
```
blegat, line 147: What do you mean by `cons` ? You mean 3 ?

docs/src/usage.md
```diff
@@ -117,4 +117,71 @@ DiffOpt.reverse_differentiate!(model)
...
+
+On the other hand, if the parameter appears in the LHS of the constraints, we can calculate the objective sensitivity with respect to the parameter using: the sensitivities of the variables with respect to the parameter, \( \frac{\partial x}{\partial p} \), and the gradient of the objective with respect to the variables \( \frac{\partial f}{\partial x} \):
```
blegat: It's not about left-hand side or right-hand side, you can also write `@constraint(model, 3 <= 2x + 3y)` in JuMP. It's about whether it is a constant term or a coefficient

docs/src/usage.md
```diff
@@ -117,4 +117,71 @@ DiffOpt.reverse_differentiate!(model)
...
+    0.1,
+)
+
+DiffOpt.reverse_differentiate!(model)
+
+MOI.get(model, DiffOpt.ReverseConstraintSet(), ParameterRef(p))
+```
+
+It is important to note that the (reverse) parameter perturbation given an objective perturbation is somewhat equivalent to the perturbation with respect to solution (since one can be calculated from the other). Therefore, one cannot set both the objective sensitivity (`DiffOpt.ReverseObjectiveSensitivity`) and the solution sensitivity (e.g. `DiffOpt.ReverseVariablePrimal`) at the same time - the code will throw an error if you try to do so.
+```
```
blegat: Remove

docs/src/usage.md
blegat: So this means you can compute it using the dual objective if you only have parameters as constraint constants and you can also use it with the primal objective using the solution sensitivity which also works even if you have parameters as constraint coefficients.
2 questions the readers might have:
1. If you only have parameters on the constraint constants, would computing the sensitivity using the primal or dual objective match ?
2. Why not use the dual objective but also include the sensitivity of the dual solution ?

blegat: Probably it would make sense to also add `ForwardDualObjectiveSensitivity` (in another PR of course) ?

docs/src/usage.md
andrewrosemberg: done

docs/src/usage.md
andrewrosemberg: good point, updated

docs/src/usage.md
andrewrosemberg: good catch

docs/src/usage.md
andrewrosemberg: This is a good point. I will add an issue for this. 

My naive response for now is:
> If you only have parameters on the constraint constants, would computing the sensitivity using the primal or dual objective match?

Yes, and we add this as a test. Also added a sentence on this:
https://github.com/jump-dev/DiffOpt.jl/blob/43d6eb86d0e0b24a9a1f2025e3390a79dbb936c7/docs/src/usage.md?plain=1#L155

> Why not use the dual objective but also include the sensitivity of the dual solution?

Added a sentence on this (and equivalence under strong duality) and will create an issue

andrewrosemberg: Sentence: https://github.com/jump-dev/DiffOpt.jl/blob/f8042b9926fed8bfaddf4bd00a6ca8854a96960c/docs/src/usage.md?plain=1#L189-L194

andrewrosemberg: Issue https://github.com/jump-dev/DiffOpt.jl/issues/315

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "DiffOpt"
 uuid = "930fe3bc-9c6b-11ea-2d94-6184641e85e7"
 authors = ["Akshay Sharma", "Mathieu Besançon", "Joaquim Dias Garcia", "Benoît Legat", "Oscar Dowson", "Andrew Rosemberg"]
-version = "0.5.1"
+version = "0.6.0"
```
blegat: Why is it breaking ?

docs/src/usage.md
```diff
@@ -117,4 +117,78 @@ DiffOpt.reverse_differentiate!(model)
...
+DiffOpt.empty_input_sensitivities!(model)
+
+MOI.set(model, DiffOpt.ForwardConstraintSet(), ParameterRef(p), Parameter(3.0))
+MOI.set(model, DiffOpt.ForwardConstraintSet(), ParameterRef(p_c), Parameter(3.0))
+DiffOpt.forward_differentiate!(model)
+
+MOI.get(model, DiffOpt.ForwardObjectiveSensitivity())
+```
+
+In the backward mode, we can calculate the parameter perturbation with respect to the objective perturbation:
```
blegat:
```suggestion
In reverse mode, we can calculate the parameter perturbation with respect to the objective perturbation:
```

Project.toml
andrewrosemberg: I thought a New, Backward-Compatible Functionality was a reason for a minor increase, but I am happy to change to 0.5.2.

andrewrosemberg:
```suggestion
version = "0.5.2"
```

andrewrosemberg: changed :)

blegat: It's a reason after v1, before v1, we only bump it for breaking changes

blegat: I still see 0.6.0 :)

andrewrosemberg:
```suggestion
version = "0.5.2"
```

andrewrosemberg: My bad, now it should habe been changed. Not sure what had happened with github apply changes.

joaquimg:
note that tests are failing

docs/src/usage.md
```diff
@@ -117,4 +117,78 @@ DiffOpt.reverse_differentiate!(model)
...
+@constraint(model, cons, pc * x >= 3 * p)
+@objective(model, Min, x^4)
+optimize!(model)
+
+direction_p = 3.0
+MOI.set(model, DiffOpt.ForwardConstraintSet(), ParameterRef(p), Parameter(direction_p))
+DiffOpt.forward_differentiate!(model)
+```
+
+Using Lagrangian duality we could already calculate the objective sensitivity with respect to parameters that appear as constants of the constraints (e.g, `cons` in this case for parameter `p`) - i.e. The objective sensitivity w.r.t. a constant parameter change is given by the optimal multiplier.
```
joaquimg:
```suggestion
Using Lagrangian duality we can easily calculate the objective sensitivity with respect to parameters that appear as constants of the constraints (e.g, `cons` in this case for parameter `p`) - i.e. The objective sensitivity w.r.t. a constant parameter change is given by the optimal dual multiplier, under strong duality.
```

docs/src/usage.md
```diff
@@ -117,4 +117,78 @@ DiffOpt.reverse_differentiate!(model)
...
+
+On the other hand, if the parameter appears as a coefficient of the constraints, one can calculate the objective sensitivity with respect to the parameter using the sensitivities of the variables with respect to the parameter, \( \frac{\partial x}{\partial p} \), and the gradient of the objective with respect to the variables \( \frac{\partial f}{\partial x} \):
+
+```math
+\frac{\partial f}{\partial p} = \frac{\partial f}{\partial x} \frac{\partial x}{\partial p}
+```
+ - A consequence of the chain-rule.
+
+Note that, if the parameter appears as a constant in a constraint, the objective sensitivity calculated through solution sensitivity is equivalent to the optimal multiplier associated with the constraint.
```
joaquimg:
```suggestion
```

this was already mentioned above

docs/src/usage.md
```diff
@@ -117,4 +117,78 @@ DiffOpt.reverse_differentiate!(model)
...
+    DiffOpt.ReverseObjectiveSensitivity(),
+    0.1,
+)
+
+DiffOpt.reverse_differentiate!(model)
+
+MOI.get(model, DiffOpt.ReverseConstraintSet(), ParameterRef(p))
+```
+
+It is important to note that the (reverse) parameter perturbation given an objective perturbation is somewhat equivalent to the perturbation with respect to solution (since one can be calculated from the other). Therefore, one cannot set both the objective sensitivity (`DiffOpt.ReverseObjectiveSensitivity`) and the solution sensitivity (e.g. `DiffOpt.ReverseVariablePrimal`) at the same time - the code will throw an error if you try to do so.
```
joaquimg, line 186: Alternatively, we could consider accumulating the two. Accumulating might be more error-prone, though.
In the end we should follow the same pattern as the case of parameters X coefficients senstivity setting: https://github.com/jump-dev/DiffOpt.jl/issues/297

docs/src/usage.md
```diff
@@ -117,4 +117,78 @@ DiffOpt.reverse_differentiate!(model)
...
+
+**Dual Objective Sensitivity**
+
+In addition to the primal objective sensitivity, one could also calculate the dual objective sensitivity with respect to the parameters using the gradients of the dual objective and the sensitivities of the dual variables with respect to the parameters.
+This is currently not implemented for any problem class, but will be available in future releases.
+
+Note that the dual objective sensitivity is equivalent to the primal objective sensitivity problems where strong duality holds.
```
joaquimg:
```suggestion
```

I would leave it for when we have it...

src/ConicProgram/ConicProgram.jl
```diff
@@ -450,4 +450,12 @@ function MOI.get(
     return MOI.get(model.model, attr, ci)
 end
 
+function MOI.get(::Model, ::DiffOpt.ForwardObjectiveSensitivity)
+    return error("Not implemented")
```
joaquimg: We should give more details here, such as: ForwardObjectiveSensitivity is not implemented for the Conic Optimization backend

src/NonLinearProgram/NonLinearProgram.jl
```diff
@@ -525,15 +526,19 @@ function DiffOpt.forward_differentiate!(model::Model; tol = 1e-6)
...
         # Compute Jacobian
-        Δs = _compute_sensitivity(model; tol = tol)
+        Δs, df_dp = _compute_sensitivity(model; tol = tol)
 
         # Extract primal and dual sensitivities
         primal_Δs = Δs[1:length(model.cache.primal_vars), :] * Δp # Exclude slacks
         dual_Δs = Δs[cache.index_duals, :] * Δp # Includes constraints and bounds
 
+        # obj sensitivity wrt parameters
+        dual_p = df_dp * Δp
```
joaquimg: I think we should rename `dual_p` to `objective_sensitivity_p`

src/NonLinearProgram/NonLinearProgram.jl
```diff
@@ -27,10 +27,11 @@ end
 Base.@kwdef struct ForwCache
     primal_Δs::Dict{MOI.VariableIndex,Float64}  # Sensitivity for primal variables (indexed by VariableIndex)
     dual_Δs::Vector{Float64}           # Sensitivity for constraints and bounds (indexed by ConstraintIndex)
+    dual_p::Float64  # Objective Sensitivity wrt parameters
```
joaquimg: see rename suggestion

src/jump_moi_overloads.jl
```diff
@@ -107,6 +111,39 @@ function MOI.get(
...
+    attr::ReverseObjectiveSensitivity,
+    val::Number,
+)
+    return MOI.set(JuMP.backend(model), attr, val)
+end
+
+function JuMP.dual(var_ref::JuMP.VariableRef; result::Int = 1)
+    JuMP.is_parameter(var_ref) || error("Variable is not a parameter")
+    return dual(ParameterRef(var_ref); result = result)
+end
```
joaquimg: This is too much type piracy.
```suggestion
```

src/jump_moi_overloads.jl
```diff
@@ -66,6 +66,10 @@ function MOI.get(
     return JuMP.jump_function(model, moi_func)
 end
 
+function MOI.get(model::JuMP.Model, attr::ForwardObjectiveSensitivity)
+    return MOI.get(JuMP.backend(model), attr)
+end
```
joaquimg: Is this really needed?

src/jump_moi_overloads.jl
```diff
@@ -107,6 +111,39 @@ function MOI.get(
...
+    return MOI.set(JuMP.backend(model), attr, JuMP.index(con_ref), val)
+end
+
+function MOI.set(
+    model::JuMP.Model,
+    attr::ReverseObjectiveSensitivity,
+    val::Number,
+)
+    return MOI.set(JuMP.backend(model), attr, val)
+end
```
joaquimg: are these really needed? JuMP should have default falbacks for these operation

src/QuadraticProgram/QuadraticProgram.jl
```diff
@@ -501,4 +501,12 @@ function MOI.set(model::Model, ::LinearAlgebraSolver, linear_solver)
     return model.linear_solver = linear_solver
 end
 
+function MOI.get(::Model, ::DiffOpt.ForwardObjectiveSensitivity)
+    return error("Not implemented")
```
joaquimg: see the conic comments

test/nlp_program.jl
```diff
@@ -627,6 +632,125 @@ function test_compute_derivatives_Finite_Diff(;
...
+    @assert is_solved_and_feasible(model)
+
+    # Set pertubations
+    Δp = 0.1
+    MOI.set(
+        model,
+        DiffOpt.ForwardConstraintSet(),
+        ParameterRef(p),
+        Parameter(Δp),
+    )
```
joaquimg: you can use: set_forward_parameter

test/nlp_program.jl
```diff
@@ -627,6 +632,125 @@ function test_compute_derivatives_Finite_Diff(;
...
+    # Parameters
+    @variable(model, p ∈ MOI.Parameter(1.5))
+    @variable(model, p_prox)
+
+    # Variables
+    @variable(model, x)
+
+    # Constraints
+    @constraint(model, con, p_prox == p) # dual fishing :)
+    @constraint(model, x * sin(p_prox) == 1)
```
joaquimg: this would not work if the parameter was directly inside the sine?

test/nlp_program.jl
```diff
@@ -627,6 +632,125 @@ function test_compute_derivatives_Finite_Diff(;
...
+        ParameterRef(p),
+        Parameter(Δp),
+    )
+
+    # Compute derivatives
+    DiffOpt.forward_differentiate!(model)
+
+    # Test Objective Sensitivity wrt parameters
+    df_dp = MOI.get(model, DiffOpt.ForwardObjectiveSensitivity())
+    @test isapprox(df_dp, dual(con) * Δp; atol = 1e-4)
```
joaquimg: It would be nice to test the actual value as well, just to make sure we are not testing a bunch of zeros...

test/nlp_program.jl
```diff
@@ -627,6 +632,125 @@ function test_compute_derivatives_Finite_Diff(;
...
+
+    # Compute derivatives
+    DiffOpt.reverse_differentiate!(model)
+
+    # Test Objective Sensitivity wrt parameters
+    dp = MOI.get(model, DiffOpt.ReverseConstraintSet(), ParameterRef(p)).value
+
+    @test isapprox(dp, dual(con) * Δf; atol = 1e-4)
+
+    # Model 2
```
joaquimg, line 698: separate test for a separate model

test/nlp_program.jl
```diff
@@ -627,6 +632,125 @@ function test_compute_derivatives_Finite_Diff(;
     end
 end
 
+################################################
+#=
+# Test Objective Sensitivity wrt Parameters
+=#
+################################################
+
+function test_ObjectiveSensitivity()
```
joaquimg: this is only testing the case with parameters in the RHS which is the one that can rely on duals.

we need some teste for parameters in the middle of nonlinear function.

my suggestion is a test with parameter * variable and another one in a non-linear term.

each of these test should have analystical solutions.

It would be nice to have a simple test where we have 10 parameters but we only differentiate with respect to the 3rd and 7th for instance. to test the sparsity handling.

joaquimg (changes requested):

src/ConicProgram/ConicProgram.jl
andrewrosemberg:
```suggestion
    return error("ForwardObjectiveSensitivity is not implemented for the Conic Optimization backend")
```

src/QuadraticProgram/QuadraticProgram.jl
andrewrosemberg:
```suggestion
    return error("ForwardObjectiveSensitivity is not implemented for the Quadratic Optimization backend")
```

src/QuadraticProgram/QuadraticProgram.jl
```diff
@@ -501,4 +501,12 @@ function MOI.set(model::Model, ::LinearAlgebraSolver, linear_solver)
     return model.linear_solver = linear_solver
 end
 
+function MOI.get(::Model, ::DiffOpt.ForwardObjectiveSensitivity)
+    return error("ForwardObjectiveSensitivity is not implemented for the Quadratic Optimization backend")
+end
+
+function MOI.set(::Model, ::DiffOpt.ReverseObjectiveSensitivity, val)
+    return error("Not implemented")
```
andrewrosemberg:
```suggestion
    return error("ReverseObjectiveSensitivity is not implemented for the Quadratic Optimization backend")
```

src/ConicProgram/ConicProgram.jl
```diff
@@ -450,4 +450,12 @@ function MOI.get(
     return MOI.get(model.model, attr, ci)
 end
 
+function MOI.get(::Model, ::DiffOpt.ForwardObjectiveSensitivity)
+    return error("ForwardObjectiveSensitivity is not implemented for the Conic Optimization backend")
+end
+
+function MOI.set(::Model, ::DiffOpt.ReverseObjectiveSensitivity, val)
+    return error("Not implemented")
```
andrewrosemberg:
```suggestion
    return error("ReverseObjectiveSensitivity is not implemented for the Conic Optimization backend")
```

src/jump_moi_overloads.jl
andrewrosemberg: If this is a must, I can remove, but there is no type piracy here right? Since dual's are not defined for variables, this is just a poetic license.

src/NonLinearProgram/NonLinearProgram.jl
andrewrosemberg: done

test/nlp_program.jl
andrewrosemberg: yes, ran this to get the results from the duals but now set it as a value test

test/nlp_program.jl
andrewrosemberg: done

test/nlp_program.jl
andrewrosemberg: done

test/nlp_program.jl
andrewrosemberg: good idea, added and changed

docs/src/usage.md
andrewrosemberg: Personal opinion is that accumulating could be problematic and hide issues for further downstream applications. At best, I think we could add an attribute that would allow for accumulating. But I am easy if you all agree otherwise.

src/ConicProgram/ConicProgram.jl
andrewrosemberg: done

src/jump_moi_overloads.jl
andrewrosemberg: removing

src/jump_moi_overloads.jl
andrewrosemberg: no, removed

src/jump_moi_overloads.jl
andrewrosemberg: removed

docs/src/usage.md
joaquimg: Agreed, lets do mutually exclusive for now

andrewrosemberg (commented):
@joaquimg Any pending change requests that I missed?

joaquimg (approved):

PR #317, opened 2025-10-05
Remove bridge instead of test

joaquimg:
If there is a follow-up to https://github.com/jump-dev/MathOptInterface.jl/pull/2863 removing the test we can remove this.
At the same time its does not seem bad to block Zeros bridge from start

### DIFF
--- a/src/moi_wrapper.jl
+++ b/src/moi_wrapper.jl
@@ -50,6 +50,19 @@ function diff_optimizer(
         with_bridge_type,
         with_cache_type,
     )
+    if !isnothing(with_bridge_type)
+        # removed because of the `ZerosBridge` issue:
+        # https://github.com/jump-dev/MathOptInterface.jl/issues/2861
+        # - zeros bridge does not support duals because it cumbersome
+        # - many bridges do not support get ConstraintFunction because it is cumbersome
+        # so there is no way out of this error for now.
+        # at the same time this is a modeling corner case tha could be avoided
+        # by the user.
+        MOI.Bridges.remove_bridge(
+            optimizer,
+            MOI.Bridges.Variable.ZerosBridge{with_bridge_type},
+        )
+    end
     if allow_parametric_opt_interface &&
        !MOI.supports_add_constrained_variable(
         isnothing(with_bridge_type) ? optimizer : optimizer.model,
--- a/test/moi_wrapper.jl
+++ b/test/moi_wrapper.jl
@@ -27,27 +27,10 @@ end
 
 function test_moi_test_runtests()
     model = DiffOpt.diff_optimizer(HiGHS.Optimizer)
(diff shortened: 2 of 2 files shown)


src/moi_wrapper.jl
```diff
@@ -50,6 +50,19 @@ function diff_optimizer(
         with_bridge_type,
         with_cache_type,
     )
+    if !isnothing(with_bridge_type)
+        # removed because of the `ZerosBridge` issue:
+        # https://github.com/jump-dev/MathOptInterface.jl/issues/2861
+        # - zeros bridge does not support duals because it cumbersome
+        # - many bridges do not support get ConstraintFunction because it is cumbersome
```
blegat, line 57: Why would we need ConstraintFunction ? I don't understand this one

blegat: This is something needed by the tests but in practice, no one asks for ConstraintFunction to the bridges outside of the tests. It's not clear in your comment if DiffOpt needs to query ConstraintFunction or not

src/moi_wrapper.jl
```diff
@@ -50,6 +50,19 @@ function diff_optimizer(
         with_bridge_type,
         with_cache_type,
     )
+    if !isnothing(with_bridge_type)
+        # removed because of the `ZerosBridge` issue:
+        # https://github.com/jump-dev/MathOptInterface.jl/issues/2861
+        # - zeros bridge does not support duals because it cumbersome
```
blegat, line 56: cumbersome ? Shouldn't we simply say that ZerosBridge does not support dual and we know with DiffOpt that we will need the duals so we can't use that bridge ?

blegat: This bridge is only useful in situations where we don't need the dual so in case we are sure in advance that we need the dual it's clear we can remove it

blegat: We can just say that in DiffOpt we know we want the duals so the case is clear

src/moi_wrapper.jl
joaquimg: DiffOpt needs a solver that supports ConstraintFuncion because it will copy those constraints from the optimizer to the .diff model

src/moi_wrapper.jl
joaquimg: One issue is, if the bridge is there, we allow the user to pass some constraints but then we disallow duals for those.
An MOI test fails and it is conceivable that someone hits this wall.
One idea is to not add that constraint, hence the removal of the bridge.

joaquimg: I will change the comment

blegat: Makes sense, then it matches what I said, we should just say that this bridge is enabled by default and MOI expects the user to remove if it is known in advance that duals will be needed. Since we know we want duals here, we know we can remove it in advance.

PR #334, opened 2026-02-02
Adds VectorNonlinearOracle support

andrewrosemberg:
Adds VectorNonlinearOracle support by creating a bridge for the nonlinear backend


andrewrosemberg:
There is only one issue. if I create a model with diff_optimizer it defaults to use conic backend even though it is not compatible with vno. Trying to figure out how to avoid this.

joaquimg (commented):
I'd like @blegat 's eyes on it.
For now, can we move the tests to seperate file only for VNO?

src/NonLinearProgram/vno_bridge.jl
```diff
@@ -0,0 +1,507 @@
...
+    end
+
+    MOI.Nonlinear.register_operator(nlm, op, n, f, ∇f, ∇²f)
+    return
+end
+
+# --------------------------------------------------------------------------
+# The actual bridge type
+# --------------------------------------------------------------------------
+struct VNOToScalarNLBridge{T<:Real} <: _B.Constraint.AbstractBridge
```
blegat, line 208: Write a short docstring

src/NonLinearProgram/vno_bridge.jl
```diff
@@ -0,0 +1,507 @@
...
+    n = b.s.input_dimension
+
+    if length(value) == m
+        # Direct mapping: value[i] is dual for output i
+        _set_duals_from_output_duals!(model, b, value)
+    elseif length(value) == n
+        # The dual is per-input-variable (J' * λ), need to recover λ
+        _set_duals_from_input_duals!(model, b, value)
+    else
+        # Unexpected size, skip
```
blegat: Throw an error here

src/NonLinearProgram/vno_bridge.jl
```diff
@@ -0,0 +1,507 @@
...
+    #
+    # For our bridged scalar constraints, we need the dual per output (λ).
+    # We can recover λ from the Jacobian: λ = (J * J')^{-1} * J * dual_per_var
+
+    m = b.s.output_dimension
+    n = b.s.input_dimension
+
+    if length(value) == m
+        # Direct mapping: value[i] is dual for output i
+        _set_duals_from_output_duals!(model, b, value)
```
blegat: If the Jacobian is square, it does not mean that it is the identity, we should still invert the Jacobian here

andrewrosemberg: A legacy from my debugging that introduced this silent bug. Thank you for catching!

joaquimg (approved):

PR #344, opened 2026-03-13
Adds support for differentable-native solvers

PTNobel:
Hey JuMP team!

I'm working on moreau.so @optimalintellect; and we natively support differentiable optimization, and can reuse some data structures between the forward and backward pass. However, to do that, we need to preserve a handle that points to the data structures inbetween the forward and backward pass.

I wanted to support DiffOpt natively in our Julia frontend, so I am adding support in DiffOpt for an MOI interface to declare that it has native differentiation support and for that to be used.

I also added documentation for this.

This is my first time using Julia, so Claude Code was invaluable. I have read all of the code, and I believe I understand it. That said, I make no promises that I have any idea what the Julia ecosystem expects, etc. so please suggest as many or as extensive of revisions as desired.


PTNobel:
The Moreau.jl package is working and passing the MOI.Test. It's waiting for approval to be made public, but right now it's very similar to SCS.jl internally

PTNobel:
I am not sure; what MOI features would that require?

src/SolverBackedDiff.jl
```diff
@@ -0,0 +1,319 @@
...
+# Unwrap the solver from DiffOpt/MOI wrapper layers
+# ─────────────────────────────────────────────────────────────────────────────
+
+"""
+    _unwrap_solver(model) -> solver_or_nothing
+
+Walk through CachingOptimizer, bridge, and POI layers to find the innermost
+solver. Returns `nothing` if no natively-differentiable solver is found.
+"""
+function _unwrap_solver(model::MOI.Utilities.CachingOptimizer)
```
blegat, line 93: We shouldn't need that, the goal of attributes is to flow through the layers so that you never need to unwrap.

blegat:
I agree with Joaquim, I don't see why we should move things to MOI. MOI has reached v1 so anything we move there is frozen. Adding DiffOpt as dependency to Moreau or even as a packages extension is preferrable.
We can just start dicussing this PR directly. DiffOpt was designed in a way to make it possible for solvers to give their derivatives directly, at the time we were thinking of doing this for OSQP and waiting for it to be part of the released version of osqp but never got to do it so I'm curious to see how to make it work with Moreau. Of course, it will be made easier once it's public but no need to wait.

I'm wondering if it wouldn't be possible to do something much simpler that what Claude Code suggested here.
I like the idea of defining
```julia
struct BackwardDifferentiate <: MOI.AbstractModelAttribute end
```
and then checking whether the solver supports it with
```julia
MOI.supports(model.optimizer, DiffOpt.BackwardDifferentiate())
```
and if it does just triggering the computation with
```julia
MOI.set(model.optimizer, DiffOpt.BackwardDifferentiate())
```
However, for setting the attributes, it shouldn't do the `_unwrap` etc... It should just pass the attributes we defined for it.

So in
https://github.com/jump-dev/DiffOpt.jl/blob/c010b536949c5eb5fe13769f444629a79f1de9af/src/moi_wrapper.jl#L551
in case `MOI.supports(model.optimizer, DiffOpt.BackwardDifferentiate())` returns `true`, then we copy the `input_cache` directly into `model.optimizer`, not `model.diff`. So maybe what we can simply do is make
https://github.com/jump-dev/DiffOpt.jl/blob/c010b536949c5eb5fe13769f444629a79f1de9af/src/moi_wrapper.jl#L785
return `model.optimizer` and `model.diff` can just be `nothing`, it won't be used.
These parameters are all passed down through the layers, and transformed accordingly (as joaquim says we implement transformation in the bridges on a per-need basis, it's easy but we only do it for a bridge when we need it, don't hesitate to let us know if it's needed for a bridge your solver is using).
Then, we request the differentiation to the solver with
```julia
MOI.set(model.optimizer, DiffOpt.BackwardDifferentiate())
```
Here, to make sure it's passed down to the solver and not the cache, maybe it should be an `AbstractOptimizerAttribute`, not an `AbstractModelAttribute`.
The, we should make
https://github.com/jump-dev/DiffOpt.jl/blob/c010b536949c5eb5fe13769f444629a79f1de9af/src/moi_wrapper.jl#L839
return `model.optimizer`, and all results will be queried from the solver as the attributes defined in DiffOpt, and transformed across the layers as expected.
This was designed with attributes for this use case to work, I just never tried it with an actual solver so there were these remaining pieces to make it work but I'd prefer to first try it this way, which should be much simpler in my opinion. If there are good reasons that this approach doesn't work then we can consider the approach of this PR (I wouldn't want Claude Code to feel offended :) )

PR #356, opened 2026-03-25
Interface for solver supporting diff natively

blegat:
First commit by Claude Code during our meeting based on this prompt https://github.com/jump-dev/DiffOpt.jl/pull/344#issuecomment-4096364587
Then I took over
Closes https://github.com/jump-dev/DiffOpt.jl/pull/344

### DIFF
--- a/src/jump_moi_overloads.jl
+++ b/src/jump_moi_overloads.jl
@@ -251,6 +251,20 @@ end
 
 MOI.constant(func::IndexMappedFunction) = MOI.constant(func.func)
 
+# Support JuMP.coefficient on plain MOI functions returned by native solvers
+function JuMP.coefficient(
+    func::MOI.ScalarAffineFunction{T},
+    vi::MOI.VariableIndex,
+) where {T}
+    coef = zero(T)
+    for term in func.terms
+        if term.variable == vi
+            coef += term.coefficient
+        end
+    end
+    return coef
+end
+
 function JuMP.coefficient(func::IndexMappedFunction, vi::MOI.VariableIndex)
     return JuMP.coefficient(func.func, func.index_map[vi])
 end
--- a/src/moi_wrapper.jl
+++ b/src/moi_wrapper.jl
@@ -550,7 +550,7 @@ MOI.get(model::Optimizer, ::ModelConstructor) = model.model_constructor
 
 MOI.supports(::Optimizer, ::ReverseDifferentiate) = true
 
-function MOI.set(model::Optimizer, attr::ReverseDifferentiate, ::Nothing)
+function MOI.set(model::Optimizer, attr::ReverseDifferentiate, value)
     st = MOI.get(model.optimizer, MOI.TerminationStatus())
     if !in(st, (MOI.LOCALLY_SOLVED, MOI.OPTIMAL))
         error(
@@ -600,7 +600,7 @@ function MOI.set(model::Optimizer, attr::ReverseDifferentiate, ::Nothing)
             end
         end
     end
-    return MOI.set(diff, attr, nothing)
+    return MOI.set(diff, attr, value)
 end
 
 # Gradient evaluation functions for objective sensitivity fallbacks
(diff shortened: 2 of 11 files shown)


test/solver_native_diff.jl
```diff
@@ -0,0 +1,1111 @@
...
+const RTOL = 1e-8
+
+# ─────────────────────────────────────────────────────────────────────────────
+# EqQPSolver: a minimal equality-constrained QP solver with native
+# differentiation support via DiffOpt.BackwardDifferentiate and
+# DiffOpt.ForwardDifferentiate.
+#
+#   min  (1/2) x'Qx + c'x
+#   s.t. Ax = b
+# ─────────────────────────────────────────────────────────────────────────────
```
odow, line 28: Did you copy this from #344?

https://github.com/jump-dev/DiffOpt.jl/pull/344/changes#diff-5c4db019bd25c07f7ae70b995aa20e622fd98ef5ba02b0d7b7b8a36307e9a742

Otherwise, this raises a lot of questions from a copyright perspective that I don't know we're well prepared to handle.

In this case it's okay. But how do we have any control that the code is coming from a JuMP-dev repo and not some external source? And even it if is from a JuMP repo, what if the license is not MIT? And how can we ensure we keep any appropriate copyright etc?

joaquimg: There must be a tool for searching online and verifying.

odow: Sure, but do I need to do that for every PR? When someone manually submits a PR we trust them that they haven't copied someone (without declaring it). I don't have that trust if it's Claude.

joaquimg: Cant we run the auto check in CI?

odow: What auto check?

odow: Oh you mean if there is such a tool, then we could run it in CI. The problem is: which tool?

blegat: Yes, it was copied from there. I agree it is a concern

odow: You copied or Claude copied?

blegat: I asked Claude to copy

blegat: Here Claude didn't take this decision by himself so it seems fine but I still think there is a risk that LLMs will steal code from places with proper attributions. They've been trained from data with licenses, disregarding this license during training anyway

blegat:
This looks good to merge, let me know if there are any comments

src/jump_moi_overloads.jl
```diff
@@ -251,6 +251,20 @@ end
 
 MOI.constant(func::IndexMappedFunction) = MOI.constant(func.func)
 
+# Support JuMP.coefficient on plain MOI functions returned by native solvers
+function JuMP.coefficient(
+    func::MOI.ScalarAffineFunction{T},
+    vi::MOI.VariableIndex,
+) where {T}
```
joaquimg, line 258: more type piracy...

src/moi_wrapper.jl
```diff
@@ -666,24 +672,30 @@ function _copy_forward_in_constraint(diff, index_map, con_map, constraints)
     return
 end
 
-function forward_differentiate!(model::Optimizer)
+MOI.supports(model::Optimizer, attr::ForwardDifferentiate) = true
+
+function MOI.set(model::Optimizer, attr::ForwardDifferentiate, value)
     st = MOI.get(model.optimizer, MOI.TerminationStatus())
```
joaquimg, line 678: Is there reasonable precedence for this?

joaquimg:
I am a bit surprise `_native_diff_solver` is not marked as covered

blegat:
> I am a bit surprise `_native_diff_solver` is not marked as covered

Ah yes, it wasn't needed at the end, it can just always use `.diff`. Removed in https://github.com/jump-dev/DiffOpt.jl/pull/359

src/moi_wrapper.jl
blegat: Using `MOI.set` to trigger a computation ? I don't think so but I think it's the most layer-friendly approach

src/jump_moi_overloads.jl
blegat: It's maybe not needed in the end, let me check

blegat: See https://github.com/jump-dev/DiffOpt.jl/pull/360

PR #372, opened 2026-07-17
Remove ProductOfSets in favor of MOI.Utilities.RuntimeProductOfSets

odow:
I wanted to update MOI (https://github.com/jump-dev/MathOptInterface.jl/pull/3022), but this part of DiffOpt depends on the internal fields.

The change to MOI is theoretically breaking (as evidenced by the need for this PR), so that is a draft for discussion. We might never merge the change in MOI, but this change could be merged regardless.


blegat:
Can the ProductOfSets in DiffOpt be simplified to be able to reuse against the implementation of the MOI one ? Actually, I always wanted to move this ProductOfSets of DiffOpt to MOI, it is also useful for MatrixOfConstraints and would be for any MOI model that can support arbitrary sets. Maybe we can move it to MOI first and then do https://github.com/jump-dev/MathOptInterface.jl/pull/3022

odow:
Maybe one required step is to compat bound DiffOpt in the registry. Then we can make changes to MOI. And then we can make changes to DiffOpt.

I'd actually be in favor of making the types in MOI settable at runtime. I don't think we gain much by having them set by the macro.

odow:
https://github.com/JuliaRegistries/General/pull/161880

blegat:
Yes, the version of DiffOpt sets them at runtime, so you mean you're in favor of moving this one to MOI ?

odow:
Updated to use https://github.com/jump-dev/MathOptInterface.jl/pull/3026

odow:
@blegat okay with this?

blegat (approved):
I've been waiting for this!

PR #377, opened 2026-08-10
Reuse the differentiation backend

joaquimg:
`_diff` is re-entered after every `optimize!` (which invalidates `model.diff`), and each rebuild instantiated a fresh `LazyBridgeOptimizer`, re-resolving a bridge graph that depends only on the problem structure.

The backend is now kept in `model.diff_backend` under a `(constructor, parametric)` key and merely `MOI.empty!`-ed on reuse, which clears model and bridge maps but preserves the resolved graph.


src/moi_wrapper.jl
```diff
@@ -777,17 +794,23 @@ function _add_bridges(instantiated_model)
...
+function _is_parametric(model::Optimizer)
     # parametric_diff = MOI.supports_constraint(
     #     model,
     #     MOI.VariableIndex,
     #     MOI.Parameter{Float64},
     # )
-    list = MOI.get(
+    # `NumberOfConstraints` rather than `ListOfConstraintIndices`: this runs on
+    # every `_diff` and only the emptiness matters, so there is no reason to
+    # allocate the whole index vector
```
joaquimg:
```suggestion
```

blegat (approved):

